PR #20, opened 2017-01-19
tweak to receiveMsgBatch

datoug:
The current implementation still has the old issue that even after timeout, more messages can be added. This patch fixes the issue.

### DIFF
--- a/services/outputhost/outputhost.go
+++ b/services/outputhost/outputhost.go
@@ -507,22 +507,16 @@ func (h *OutputHost) ReceiveMessageBatch(ctx thrift.Context, request *cherami.Re
 	moreResultTimer := common.NewTimer(timeoutTime.Sub(time.Now()))
 	defer moreResultTimer.Stop()
 MORERESULTLOOP:
-	for remaining := count - 1; remaining > 0; {
+	for remaining := count - 1; remaining > 0; remaining-- {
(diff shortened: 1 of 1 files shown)


services/outputhost/outputhost.go
```diff
@@ -507,22 +507,16 @@ func (h *OutputHost) ReceiveMessageBatch(ctx thrift.Context, request *cherami.Re
...
 		select {
-		case msg := <-cgCache.msgsRedeliveryCh:
-			deliver(msg, cgCache.msgCacheRedeliveredCh)
-			remaining--
-		case msg := <-cgCache.msgsCh:
-			deliver(msg, cgCache.msgCacheCh)
-			remaining--
+		case <-moreResultTimer.C:
+			break MORERESULTLOOP
 		default:
```
aravindvs: then i don't think the "default" makes sense here.. let's get rid of that one as well.

datoug: which 'default'? Line 518?
I think that's still needed. So in each loop we first check whether it's timed out. Only if not time out(default case), we check both message availability and the timer.

aravindvs: Oh.. sorry I missed it.. i thought you had all three cases on both places.. looks good now!

PR #23, opened 2017-01-19
Make Extent targets configurable; add prefix matching config rules

GuillaumeBailey:


common/util.go
```diff
@@ -605,3 +605,72 @@ func NewCliHelper() CliHelper {
...
+				hasMatch = true
+			}
+		}
+	}
+
+	if hasMatch { 
+		if len(longestMatchKey) > 1 { // Don't log for very short prefixes, e.g. '/', '' 
+			// Log just once for this particular rule; if the value or rule changes, log again
+			logMapKey := valName + path + longestMatchKey + strconv.Itoa(int(longestMatchValue))
+			if _, ok := overrideValueByPrefixLogMap[logMapKey]; !ok {
```
venkat1109: do you really need this optimization ? Instead, can you just make the log debug ? Then the log level can be changed by the user if needed to see this log. If it impacts perf, we can always change it later.

common/util.go
```diff
@@ -605,3 +605,72 @@ func NewCliHelper() CliHelper {
...
+				overrideValueByPrefixLogMap[logMapKey] = struct{}{}
```
venkat1109: Access to map not thread safe here.

common/util_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+	logFnInvocations = 0
+	for i := 1; i < 4; i++ {
+		x = OverrideValueByPrefix(logFn, `/foo/bar`, badRules, 42, `3TestOverrideValueByPrefixNil`)
+		s.EqualValues(x, 42)
+		s.Equal(logFnInvocations, i*5)
+	}
+}
+
+func (s *UtilSuite) TestOverrideValueByPrefixNormal() {
```
venkat1109: Would be cool to have one of these tests run the rule evaluations concurrently i.e. spin go routines inside your for loop. That would have caught the concurrent unsafe access to map.

services/controllerhost/api_handlers.go
```diff
@@ -41,12 +41,12 @@ const (
...
 
 const (
-	minOpenExtentsForDstPlain                    = 4
-	maxExtentsToConsumeForDstPlain               = 8
-	extentsToConsumePerRemoteZone                = 4
 	minOpenExtentsForDstDLQ                      = 1
 	maxExtentsToConsumeForDstDLQ                 = 2
 	minOpenExtentsForDstTimer                    = 2
+	defaultMinOpenExtents                        = 2 // Only used if the extent configuration can't be retrieved
+	defaultRemoteExtents                         = 2
```
venkat1109: naming nit: Can you make this const name specific ? its not clear from the name, if this is a number or a list of extents. Also, is this for publish/consume or both ?

services/controllerhost/config.go
```diff
@@ -49,15 +49,22 @@ type (
 	OutputPlacementConfig struct {
 		AdminStatus string `name:"adminStatus" default:"enabled"`
 	}
+
+	ExtentAssignmentConfig struct {
```
venkat1109: This should really be ControllerDynamicConfig or simply ControllerConfig. When we start implementing other controller features (like placement / auto-scaling), those config params would also be part of this same config struct. The idea is to have one per service.

services/controllerhost/config.go
```diff
@@ -49,15 +49,22 @@ type (
...
+		PublishExtentTargets       []string `name:"publishExtentTargets" default:"/=4"`
```
venkat1109: naming nits: Please make it clear that the config values here are numbers, `Targets` are confusing to me. One suggestion would be: 

    NumPublisherExtentsByPath []string
    NumConsumerExtentsByPath []string
    NumRemoteConsumerExtentsByPath []string

services/controllerhost/consumer.go
```diff
@@ -68,27 +68,51 @@ func newCGExtentsByCategory() *cgExtentsByCategory {
...
+				totalZones++
+			}
+		}
+		remoteZones = common.MaxInt(0, totalZones-1)
+	}
+
+	cfgIface, err := context.cfgMgr.Get(common.ControllerServiceName, `*`, `*`, `*`)
+	if err != nil {
+		logFn().WithFields(bark.Fields{common.TagErr: err}).Error(`Couldn't get extent target configuration`)
+		return defaultMinOpenExtents * 2
```
venkat1109: Use the `defaultMinConsumeExtents` constant here

services/controllerhost/api_handlers.go
```diff
@@ -55,7 +55,7 @@ const (
 // limit on how many new extents we make available
 // for consumption to a consumer group in a single
 // call to GetOutputHosts
-const maxNewExtentsForCGPerCall = 4
+const maxNewExtentsForCGPerCallFactorPerMille = 500
```
venkat1109: I believe this const is unused now, you can just delete this.

common/util.go
GuillaumeBailey: Which optimization? not logging for short rules, or not logging more than once? Not logging more than once is definitely needed, or this will be the new king of log spam. The other one can be removed, but you'll get a listing of every destination and consumer group in the log when the controller starts.

venkat1109: Both. Its just not the right place and/or way to solve the 'get visibility without log noise' problem IMO. What if I need visibility for other service config params elsewhere in the code ? Reimplement this pattern ? Some thoughts on getting this visibility :

* common/dconfig/configmgr.go does a periodic refresh and caches the result. Every time a refresh happens, we can compare the config to the old version and if there is a diff, log the whole config struct once there.
* expose a http verb in controller to dump the config for a given key
* implement a ThrottledLogger wrapper around logger that only logs 1 out of N messages

Will leave it up to you, if you still want to go down this road, I am ok with not logging the short rules.

common/util.go
```diff
@@ -605,3 +605,72 @@ func NewCliHelper() CliHelper {
...
 	}
 }
+
+var overrideValueByPrefixLogMap = make(map[string]struct{})
+
+// OverrideValueByPrefix takes a list of override rules in the form 'prefix=val' and a given string, and determines the most specific rule
+// that applies to the given string. It then replaces the given default value with the override value. logFn is a logging closure that
+// allows lazy instatiation of a logger interface to log error conditions and override status. valName is used for logging purposes, to
+// differentiate multiple instantiations in the same context
+func OverrideValueByPrefix(logFn func() bark.Logger, path string, overrides []string, defaultVal int64, valName string) int64 {
```
GuillaumeBailey: I looked at both of those places, and I don't like them for this. All the functions in those areas are receivers, and this doesn't seem like it should be one.

The function is pretty generic, it could be used to apply prefix rules for dconfig, yaml, CLI, etc.

I'd like to leave it here.

common/util.go
```diff
@@ -605,3 +605,72 @@ func NewCliHelper() CliHelper {
 		cZones:            make(map[string]string),
 	}
 }
+
+var overrideValueByPrefixLogMap = make(map[string]struct{})
+
+// OverrideValueByPrefix takes a list of override rules in the form 'prefix=val' and a given string, and determines the most specific rule
+// that applies to the given string. It then replaces the given default value with the override value. logFn is a logging closure that
+// allows lazy instatiation of a logger interface to log error conditions and override status. valName is used for logging purposes, to
+// differentiate multiple instantiations in the same context
```
GuillaumeBailey: Added.

common/util.go
```diff
@@ -605,3 +605,72 @@ func NewCliHelper() CliHelper {
...
+		if len(split) != 2 {
+			logFn().WithFields(bark.Fields{`rule`: ovrd, `valName`: valName}).Error(`Invalid override rule, couldn't split`)
+			continue moreOverrides
+		}
+		if strings.HasPrefix(path, split[0]) {
+			if len(split[0]) > len(longestMatchKey) || (split[0] == `` && !hasMatch) { // Match for a longer key, or just the empty default
+				var lmv int
+				lmv, err = strconv.Atoi(split[1])
+				if err != nil {
+					logFn().WithFields(bark.Fields{`rule`: ovrd, `valName`: valName, TagErr: err}).Error(`Invalid override rule, couldn't covert value`)
```
GuillaumeBailey: done

common/util.go
GuillaumeBailey: These options are all rather heavy weight. I'd just like to be able to say that destination X has extent config Y, based on ELK or similar, and without logging every time. 

Regarding other config params, I'd only be concerned with logging them if they have some kind of run-time interpretation like this does.

common/util.go
GuillaumeBailey: fixed.

common/util_test.go
GuillaumeBailey: done.

services/controllerhost/config.go
GuillaumeBailey: done.

services/controllerhost/api_handlers.go
GuillaumeBailey: done.

services/controllerhost/config.go
GuillaumeBailey: done

services/controllerhost/consumer.go
GuillaumeBailey: done

common/util.go
```diff
@@ -605,3 +605,82 @@ func NewCliHelper() CliHelper {
 		cZones:            make(map[string]string),
 	}
 }
+
+var overrideValueByPrefixLogMapLock sync.Mutex
```
venkat1109: how about a RWMutex ?

GuillaumeBailey: I looked at it, but you can't hold the r lock and then take the w lock, so the implementation is a bit ugly (have to check the map twice, once under r lock, second under w lock). I don't expect this to be high contention.

venkat1109: This API is expected to be called for every GetInputHosts() call, disagree that this is low contention. You are right that you have to check the map twice (on a miss), but that's ok.

PR #26, opened 2017-01-20
glide up

GuillaumeBailey:


venkat1109 (approved):

glide.lock
```diff
@@ -123,7 +121,7 @@ imports:
   - .generated/go/shared
   - .generated/go/store
 - name: github.com/uber/ringpop-go
-  version: 8b1085ca9b31007b996e411f2cb88256b2fedd12
```
aravindvs: I am not sure why glide up updated this revision.. we should be sticking to 0.7.0 which we specify in glide.yaml.. 
This is a breaking change in ringpop and we should not update it..

aravindvs (approved):
Looks great! Thanks for fixing the version.

PR #31, opened 2017-01-23
Dockerfile for standalone cherami-server

thuningxu:
This PR contains files to build and run a docker container to run Cherami in single-server, standalone mode. The container includes Cassandra and all the Cherami binaries. The image is currently public at Docker registry as `thuningxu/cherami-server:latest`.

To start the container with name "cherami":
`$ docker run -d --name cherami thuningxu/cherami-server-standalone:latest`

Please wait about 30 seconds for the Cassandra and Cherami to start up. Then, you can connect to it using the container IP address:

`$ docker exec cherami hostname --ip-address`
`172.17.0.2`

You can use the tools inside the container to interact with Cherami, for example, create a destination:
`$ docker exec cherami ./cherami-cli --env=prod --hostport=172.17.0.2:4922 c d "/test/cherami"`

Create a consumer group:
`$ docker exec cherami ./cherami-cli --env=prod --hostport=172.17.0.2:4922 c c "/test/cherami" "/test/cherami_reader"`

Publish from terminal (need `-it` flag for docker):
`$ docker exec -it cherami ./cherami-cli --env=prod --hostport=172.17.0.2:4922 p "/test/cherami"`

Consume:
`$ docker exec cherami ./cherami-cli --env=prod --hostport=172.17.0.2:4922 r "/test/cherami" "/test/cherami_reader"`

### DIFF
--- a/docker/server/Dockerfile
+++ b/docker/server/Dockerfile
@@ -1,0 +1,35 @@
+# Copyright (c) 2016 Uber Technologies, Inc.
+
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+FROM golang:1.7.4
+
+# get and compile cherami-server
+ENV CHERAMI_HOME $GOPATH/src/github.com/uber/cherami-server
+
+RUN go get -u github.com/Masterminds/glide
+RUN git clone --depth=50 https://github.com/uber/cherami-server.git $CHERAMI_HOME
+RUN cd $CHERAMI_HOME; make bins
+
+EXPOSE 4922 4253 4240 4254 5425 6280 6189 6190 6191 6310
+
+WORKDIR $CHERAMI_HOME
+
+# Remember to mount host directories (config and data), and have CHERAMI_CONFIG_DIR properly point to config.
+CMD ["./cherami-server", "start", "all"]
--- a/docker/standalone/Dockerfile
+++ b/docker/standalone/Dockerfile
@@ -1,0 +1,63 @@
+# Copyright (c) 2016 Uber Technologies, Inc.
+
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+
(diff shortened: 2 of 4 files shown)


docker/entrypoint.sh
```diff
@@ -0,0 +1,40 @@
...
+# start up cassandra
+pushd /
+./docker-entrypoint.sh cassandra
+popd
+sleep 20
+
+export HOST_IP=`hostname --ip-address`
+
+# setup schema
+sed -i.bak "s/'replication_factor':\ '5'/'replication_factor':\ '1'/g" clients/metadata/schema/metadata_keyspace.cql
```
aravindvs: Can the RF be a config option?

aravindvs:
This is great! I am wondering if we need to expose the "RingHosts" config as part of the template.yaml as well? That way we can use the same mechanism to form a cherami cluster right? i.e, modify the RingHosts to include the ip address of all your machines (or a subset of your machines that are considered a seed) where you will be running cherami and start the docker container on all machines?

docker/entrypoint.sh
```diff
@@ -0,0 +1,40 @@
...
+ln -sf `which cqlsh` /usr/local/bin/cqlsh
+CQLSH_HOST=$HOST_IP ./scripts/cherami-setup-schema
```
venkat1109: where is this command used ?

docker/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+FROM cassandra:3.9
+
+# get golang 1.7.4 (https://github.com/docker-library/golang/blob/18ee81a2ec649dd7b3d5126b24eef86bc9c86d80/1.7/Dockerfile)
+RUN apt-get update && apt-get install -y --no-install-recommends \
```
venkat1109: Would be cool to have a macos version as well.. may be later.

docker/entrypoint.sh
```diff
@@ -0,0 +1,40 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+# start up cassandra
+pushd /
+./docker-entrypoint.sh cassandra
```
venkat1109: where is this file ?

docker/Dockerfile
thuningxu: This works under MacOS. You need to install Docker for Mac.

docker/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+VOLUME /var/lib/cherami-store
+EXPOSE 4922 4253 4240 4254 5425 6280 6189 6190 6191 6310
+
+COPY ./entrypoint.sh $CHERAMI_HOME/entrypoint.sh
+COPY ./docker_template.yaml $CHERAMI_HOME/config/docker_template.yaml
+RUN chmod a+x $CHERAMI_HOME/entrypoint.sh
+
+WORKDIR $CHERAMI_HOME
+ENTRYPOINT $CHERAMI_HOME/entrypoint.sh
+CMD ./cherami-server start all
```
thuningxu: I am still not quite clear about ENTRYPOINT vs CMD.

docker/entrypoint.sh
thuningxu: This image inherits "cassandra:3.9" (`FROM cassandra:3.9`). `docker-entrypoint.sh` is from that image.

docker/entrypoint.sh
thuningxu: which command?

docker/entrypoint.sh
thuningxu: Not sure what's the right way to do that. Maybe redo ./scripts/cherami-setup-schema to accept parameters. In that case it should be in a different PR.

thuningxu:
Regarding making RingHosts configurable: this particular image is for single instance, as it also includes Cassandra inside. For cluster deployment I would make a different image that contains Cherami only.

docker/Dockerfile
venkat1109: ah, yes.. don't know what I was thinking before..

docker/entrypoint.sh
venkat1109: Question was about CQLSH_HOST.. Forgot it was an env that cqlsh understands. nvm..

thuningxu:
Added new image "thuningxu/cherami-server" that is suitable for cluster deployment. It does not contain Cassandra, and needs externally supplied config.

PR #32, opened 2017-01-23
Output leak fixes

aravindvs:
This is the corresponding equivalent of the input leak fixes. Stress tests over the weekend looks good. 

This makes sure:
(1) we cleanup all connections/extents properly by making sure we wait appropriately before breaking the manageRoutine
(2) make sure all the buffered channels are closed
(3) make sure the message cache entries are cleaned up during shutdown so that we don't keep any existing references.

The only remaining leak is this known leak with maps in go which will be fixed in 1.8:
https://github.com/golang/go/issues/16070

### DIFF
--- a/services/outputhost/cgcache.go
+++ b/services/outputhost/cgcache.go
@@ -408,7 +408,6 @@ func (cgCache *consumerGroupCache) manageConsumerGroupCache() {
 			}
 			cgCache.extMutex.Unlock()
 		case <-cgCache.closeChannel:
-			cgCache.logger.Info("stopped")
 			// stop the delivery cache as well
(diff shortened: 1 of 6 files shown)


aravindvs:
@datoug - Assigning this to you since you just reviewed the input side last week.

services/outputhost/cgcache.go
```diff
@@ -405,6 +411,11 @@ func (cgCache *consumerGroupCache) manageConsumerGroupCache() {
 			cgCache.logger.Info("stopped")
```
datoug: Since we're doing a bunch of stuff here, probably useful to add a log line after everything is done

PR #38, opened 2017-01-24
fix a busy loop in replicator which is causing high CPU usage

datoug:
This default case is causing a busy loop when there's no message available.


services/replicator/inconn.go
```diff
@@ -150,9 +150,6 @@ func (conn *inConnection) writeMsgsStream() {
 				return
 			}
 		} else {
-			// Note: we check closeChannel only in the default case.
```
datoug: this is not a concern any more since we don't close the channel now in replicator (store will close the channel when seal marker is read)

PR #39, opened 2017-01-25
Make message cache size configurable per CG

aravindvs:
This patch makes sure we can configure the message cache size per
CG and has the following pieces:
(1) Move the cassandra based config manager behind the interface so that we can easily write tests
(2) Update existing code which got the hard type to get the interface
(3) Modify outputhost to get the new message cache size value from configMgr. Update tests to use this value.

### DIFF
--- a/services/outputhost/cgcache.go
+++ b/services/outputhost/cgcache.go
@@ -28,6 +28,7 @@ import (
 	"github.com/uber/tchannel-go/thrift"
 
 	"github.com/uber/cherami-server/common"
+	"github.com/uber/cherami-server/common/dconfig"
 	"github.com/uber/cherami-server/common/metrics"
 	"github.com/uber/cherami-server/services/outputhost/load"
 	"github.com/uber/cherami-thrift/.generated/go/cherami"
@@ -168,6 +169,7 @@ type (
 
 		// hostMetrics represents the host level load metrics reported to controller
 		hostMetrics *load.HostMetrics
+
 		// cgMetrics represents consgroup level load metrics reported to controller
 		cgMetrics *load.CGMetrics
 
@@ -179,6 +181,9 @@ type (
 
 		// manageMsgCacheWG is used to wait for the manage delivery cache routine to go away. This is needed to make sure we cleanup all the buffered channels appropriately.
 		manageMsgCacheWG sync.WaitGroup
+
+		// cfgMgr is the reference to the cassandra backed cfgMgr
+		cfgMgr *dconfig.CassandraConfigManager
 	}
 )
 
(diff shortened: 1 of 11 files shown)


services/outputhost/cgcache.go
```diff
@@ -308,6 +317,11 @@ func (cgCache *consumerGroupCache) loadExtentCache(destType shared.DestinationTy
 		cgCache.cgMetrics.Increment(load.CGMetricNumOpenExtents)
 		cgCache.hostMetrics.Increment(load.HostMetricNumOpenExtents)
 
+		// get the initial credits based on the message cache size
+		cfg, err := cgCache.getDynamicCgConfig()
+		if err == nil {
+			extCache.initialCredits = cgCache.getMessageCacheSize(cfg, defaultNumOutstandingMsgs)
```
GuillaumeBailey: We need to set extCache.initialCredits to a default value if the err != nil, right?

aravindvs: @GuillaumeBailey it is set already above in line 312 (already existing code).

services/outputhost/config.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/uber/cherami-server/common"
+	"github.com/uber/cherami-server/common/dconfig"
+	m "github.com/uber/cherami-thrift/.generated/go/metadata"
+)
+
+type (
+	// OutputCgConfig is the per cg config used by the
+	// cassandra config manager
+	OutputCgConfig struct {
+		MessageCacheSize []string `name:"messagecachesize" default:"/=10000"`
```
venkat1109: can you add a comment stating that each entry in the slice is a tuple or change the name of the member to reflect that.

services/outputhost/messagecache.go
```diff
@@ -781,6 +786,8 @@ func (msgCache *cgMsgCache) manageMessageDeliveryCache() {
 					msgCache.numAcks = 0
 				}
 			}
+		case <-msgCache.cfgRefreshTicker.C:
```
venkat1109: Do you need another ticker ? Can you do this refresh whenever redeliveryTicker fires ? Since the configMgr already caches the object, calling Get() should be light-weight.

services/outputhost/messagecache.go
```diff
@@ -781,6 +786,8 @@ func (msgCache *cgMsgCache) manageMessageDeliveryCache() {
...
+			msgCache.refreshCgConfig(msgCache.maxOutstandingMsgs)
```
venkat1109: The creditBatchSize (declared above) is derived out of this maxOutstandingMsgs, so when this gets refreshed, you likely also want to adjust the credit batch size to the store.

services/outputhost/messagecache.go
aravindvs: Yeah I was mainly worried about the calls to cassandra every 500ms (which is the redeliveryTicker) but since it is caching it, i will update to use the same ticker..

services/outputhost/messagecache.go
aravindvs: Good catch! will fix it.

services/outputhost/config.go
aravindvs: forgot to add this.. will add it now!

PR #42, opened 2017-01-26
Add test hooks to metrics SimpleReporter

GuillaumeBailey:
This pull request allows:

* Testing of metrics
* Tests that use the values of metrics

It does this by:

* Defining a type of handler function that can be implemented by a test and called whenever a metric value changes
* Allowing registration of these functions
* Implementing filtering rules to allow a handler to be triggered for only a subset of metrics updates, as desired.

### DIFF
--- a/common/metrics/testreporter.go
+++ b/common/metrics/testreporter.go
@@ -1,0 +1,194 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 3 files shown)


thuningxu (changes requested):
I think we should make a `TestReporter` instead of adding this logic to `SimpleReporter`.

common/metrics/simplereporter.go
```diff
@@ -36,6 +40,11 @@ type (
 	}
 )
 
+type HandlerFn func(metricName string, baseTags, tags map[string]string, value int64)
+
+var handlers = make(map[string]map[string]HandlerFn) // Key1 - metricName; Key2 - "filterTag:filterVal"
+var handlerMutex sync.RWMutex
+
```
thuningxu: Shouldn't `handlers` and `handlerMutex` be part of `SimpleReporter` struct? If someone creates two `SimpleReporter` instances the handlers will step on each other.

GuillaumeBailey: This is entirely by-design. Since we make the parent reporters as part of test setup, and then the product code makes sub-reporters and sub-sub-reporters, associating the handlers would require either defining the handler during test setup (and propagating to all children) or  having additional test hooks in the product to get a reference to particular child reporter.

GuillaumeBailey:
> I think we should make a TestReporter instead of adding this logic to SimpleReporter.

The TestReporter would simply be a copy of the SimpleReporter, and the SimpleReporter would thereafter be unused, which creates code duplication and dead code. 

There's very low perf impact to these changes, as the metric reporting acquires read locks only, and no processing occurs other than verifying that no handlers are registered.

This modified simpleReporter still achieves the goal of having a no-op metrics reporter.

thuningxu (approved):

PR #44, opened 2017-01-27
Smart retry auto disable

GuillaumeBailey:
This change adds product code and tests to automatically disable smart retry when a DLQ merge is in progress.

Key pieces:
- Change to cgCache to lookup single CG visibility, and calculate whether DLQ is merging
- Change to isStalled to disable smart retry when the dlqMerging flag is set
- Test that verifies a complete dlq merge cycle, runtime 70s (100s w/ race)
- Small fix on consConnection to fix metrics spamming that I noticed during testing
- Fix for a race in logPumpHealth

### DIFF
--- a/common/metrics/simplereporter.go
+++ b/common/metrics/simplereporter.go
@@ -20,7 +20,11 @@
 
 package metrics
 
-import "time"
+import (
+	"fmt"
+	"sync"
+	"time"
+)
 
 type (
 	// SimpleReporter is the reporter used to dump metric to console for stress runs
@@ -35,6 +39,11 @@ type (
 		elasped    time.Duration
 	}
 )
+
+type HandlerFn func(metricName string, baseTags, tags map[string]string, value int64)
+
+var handlers = make(map[string]map[string]HandlerFn) // Key1 - metricName; Key2 - "filterTag:filterVal"
+var handlerMutex sync.RWMutex
 
 // NewSimpleReporter create an instance of Reporter which can be used for driver to emit metric to console
 func NewSimpleReporter(tags map[string]string) Reporter {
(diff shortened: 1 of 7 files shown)


services/outputhost/cgcache.go
```diff
@@ -500,9 +502,55 @@ func (cgCache *consumerGroupCache) refreshCgCache(ctx thrift.Context) error {
 		}
 	}
 
+	newDLQMerging := nSingleCGExtents > 0
+	if cgCache.dlqMerging != newDLQMerging {
+		cgCache.dlqMerging = newDLQMerging
+		if newDLQMerging {
+			cgCache.logger.WithField(`nSingleCGExtents`, nSingleCGExtents).Info("Merging DLQ Extent(s)")
```
venkat1109: s/Merging/Serving.. Just to be clear

services/outputhost/cgcache.go
```diff
@@ -500,9 +502,55 @@ func (cgCache *consumerGroupCache) refreshCgCache(ctx thrift.Context) error {
...
+
+		} else {
+			cgCache.logger.Info("Done merging DLQ Extent(s)")
```
venkat1109: same as above.

services/outputhost/cgcache.go
```diff
@@ -500,9 +502,55 @@ func (cgCache *consumerGroupCache) refreshCgCache(ctx thrift.Context) error {
...
+		}
+	}
+
 	return nil
 }
 
+var checkSingleCGVisibleCacheLock sync.RWMutex
+var checkSingleCGVisibleCache = make(map[string]bool)
```
venkat1109: Can you move these global static vars into cgCache ? Also don't like the fact that this cache has no eviction policy. Consider using the LruCache in common.

services/outputhost/cgcache.go
```diff
@@ -500,9 +502,55 @@ func (cgCache *consumerGroupCache) refreshCgCache(ctx thrift.Context) error {
...
+	writeFn := func() {
+		req := &metadata.ReadExtentStatsRequest{
+			DestinationUUID: common.StringPtr(cgCache.cachedCGDesc.GetDestinationUUID()),
+			ExtentUUID:      common.StringPtr(cge.GetExtentUUID()),
+		}
+		ext, err := cgCache.metaClient.ReadExtentStats(ctx, req)
+		if err == nil {
+			if ext != nil && ext.GetExtentStats() != nil {
+				val = ext.GetExtentStats().GetConsumerGroupVisibility() != ``
+				if val {
```
venkat1109: you also want to update your cache/map here.

services/outputhost/cgcache.go
```diff
@@ -486,11 +489,10 @@ func (cgCache *consumerGroupCache) refreshCgCache(ctx thrift.Context) error {
...
 		// Now we have the cgCache. check and load extent for all extents
 		for _, cge := range cgRes.GetExtents() {
-			nExtents++
-			errR := cgCache.loadExtentCache(dstDesc.GetType(), cge)
-			if errR != nil {
-				return errR
+			if cgCache.checkSingleCGVisible(ctx, cge) {
+				nSingleCGExtents++
 			}
+			cgCache.loadExtentCache(dstDesc.GetType(), cge)
```
venkat1109: you are now ignoring the error from loadExtentCache (previously not). If loadExtentCache() is never supposed to fail, change the function signature to not return an error.

services/outputhost/messagecache.go
```diff
@@ -1069,9 +1070,11 @@ func (msgCache *cgMsgCache) isStalled() bool {
 	var m3St m3HealthState
 	var smartRetryDisabled bool
 
-	if strings.Contains(msgCache.GetOwnerEmail(), SmartRetryDisableString) {
+	msgCache.cgCache.extMutex.RLock()
```
venkat1109: is the lock necessary to protect access to `dlqMerging` ? If so, can you make it an atomic int instead ?

services/outputhost/messagecache.go
```diff
@@ -1206,12 +1212,10 @@ func (msgCache *cgMsgCache) logStateMachineHealth() {
 	}).Info(`state machine health`)
 }
 
-var logPumpHealthOnce bool
+var logPumpHealthOnce sync.Once
```
venkat1109: prefer moving this inside of the cgMsgCache struct.

common/metrics/simplereporter.go
```diff
@@ -78,12 +87,78 @@ func (r *SimpleReporter) GetTags() map[string]string {
...
+// the name/value of one of the metric's tags. If the filterTag/Val are both empty, all updates to that metric will
+// trigger the handler. If metricName is empty, all metrics matching the tag filter will pass through your function.
+// A nil handler unregisters the handler for the given filter parameters
+//
+// Dev notes:
+// * It is advisible to defer a call to unregister your handler when your test ends
+// * Your handler can be called concurrently. Capture your own sync.Mutex if you must serialize
+// * Counters report the delta; you must maintain the cumulative value of your counter if it is important
+// * Your handler executes synchronously with the metrics code; DO NOT BLOCK
+func RegisterHandler(metricName, filterTag, filterTagVal string, handler HandlerFn) {
```
venkat1109: Xu had a comment in another PR about implementing this logic just for the testReporter. Agree with Xu here, can we move this logic into an alternative reporter implementation, just for test ?

common/metrics/simplereporter.go
```diff
@@ -36,6 +40,11 @@ type (
 	}
 )
 
+type HandlerFn func(metricName string, baseTags, tags map[string]string, value int64)
+
+var handlers = make(map[string]map[string]HandlerFn) // Key1 - metricName; Key2 - "filterTag:filterVal"
```
venkat1109: move this inside the reporter struct

services/outputhost/cgcache.go
GuillaumeBailey: If you scroll over, you'll see that I did remove the error return, which was always nil

services/outputhost/cgcache.go
GuillaumeBailey: The feature is called 'DLQ merge'. Someone looking for code related to merging will want to look at this.

services/outputhost/cgcache.go
venkat1109: Oh, cool ! I missed that.

services/outputhost/cgcache.go
GuillaumeBailey: done

services/outputhost/cgcache.go
GuillaumeBailey: good catch

services/outputhost/messagecache.go
GuillaumeBailey: done

services/outputhost/messagecache.go
GuillaumeBailey: This one is meant to be global; when we have different capacities per cgMsgCache, then this will make sense.

GuillaumeBailey:
> Would be cool to also expose a dynamic config turn on/off knob for the new feature (as a rollback plan).

Trying to come up with a scenario where this knob would be used is hurting my head. The feature is only active during DLQ merge, which is already quite rare. Using the knob in that situation means that smart retry would be enabled again, which really only makes sense if the customer is 100% sure that all DLQ messages have been properly handled. If they are so attentive, why not just take their group offline when their downstream is down while merging?

venkat1109:
>> Would be cool to also expose a dynamic config turn on/off knob for the new feature (as a rollback plan).

Less worried about the smartRetryDisable during merge; more worried about the additional metadata ReadExtentStats call on every refresh for every CG. Will leave it up to you, it was just a question / suggestion to start the discussion around rollback. Even if we do implement a dynamic config switch, it likely will be temporary (just for rollback).

GuillaumeBailey:
> why can't we make a metdata change to flag an extent in the CGE table as a dlq extent?

We absolutely can. The cost will be:

* One change to metadata to add the singleCGVisible flag to the consumer group extents table
* Change to metadata client to store the new flag
* Change to metadata test to test the new flag
* One change to thrift to add the new field to the result consumer group extent
* Changes to test to cover the cases around the new field
* Change to controller to forward this information
* Changes to controller test to test the forwarding
* Similar changes to what already exists in outputhost to use the new information
* Similar testing for outputhost
* Rollback plan(s) for both thrift and metadata changes
* Rollback test(s) for both thrift and metadata changes

That will allow us to save:

* ~30 lines of code in outputhost
* ~one listextentstats call per extent

I didn't feel the cost was justified for the benefits

aravindvs:
Agree with most of your assessment of the no: of small changes everywhere. The reason why I was asking was because we are performing a ReadExtentStats() metadata call (albeit just once per cge) which could introduce another metadata delay (we already are seeing an increased metadata latency in prod).. We can run this and find out the perf impact and determine what to do as well. 

Although we need to address the metadata latency issue on it's own, (there are already 900 cuts) I was just wondering if we can avoid adding another cut to it :)

PR #45, opened 2017-01-28
add validation on multi_zone destination creation, include zone configs in list and show destination output

datoug:


tools/common/lib.go
```diff
@@ -228,6 +231,10 @@ func CreateDestination(c *cli.Context, cClient ccli.Client, cliHelper common.Cli
 		ExitIfError(errors.New(strWrongZoneConfigCount))
 	}
 
+	if len(zoneConfigs.Configs) > 1 && unconsumedMessagesRetention < MinUnconsumedMessagesRetentionForMultiZoneDest {
```
venkat1109: Can you do the same check within updateDestination() ? Also a comment for why this is necessary would be helpful.

datoug:
@venkat1109 I noticed an existing bug in updateDestination and fixed it in this PR. Could you take another look?

services/frontendhost/frontend.go
```diff
@@ -238,11 +238,21 @@ func convertCreateDestRequestToInternal(createRequest *c.CreateDestinationReques
 func convertUpdateDestRequestToInternal(updateRequest *c.UpdateDestinationRequest, destUUID string) *shared.UpdateDestinationRequest {
 	internalUpdateRequest := shared.NewUpdateDestinationRequest()
 	internalUpdateRequest.DestinationUUID = common.StringPtr(destUUID)
-	internalUpdateRequest.Status = common.InternalDestinationStatusPtr(shared.DestinationStatus(updateRequest.GetStatus()))
```
datoug: this function is causing the bug. Basically if the field is not specified in CLI, we still incorrectly created a pointer, then in metadata_cassandra it'll treat this field as 'specified' and update metadata using these default values.

venkat1109:
Looks good to me !

PR #47, opened 2017-01-30
MetadataMgr: Move back to controllerhost

venkat1109:
This patch moves the MetadataMgr interface & impl within the controller (where it originally belonged).  Now the controller implementation is just a decorator around the common metadata client (that emits metrics).

### DIFF
--- a/services/controllerhost/controllerhost.go
+++ b/services/controllerhost/controllerhost.go
@@ -43,6 +43,7 @@ import (
 	"github.com/uber/cherami-server/common"
 	"github.com/uber/cherami-server/common/configure"
 	"github.com/uber/cherami-server/common/dconfig"
+	"github.com/uber/cherami-server/common/metadata"
 	"github.com/uber/cherami-server/common/metrics"
 	"github.com/uber/cherami-server/services/controllerhost/load"
 	a "github.com/uber/cherami-thrift/.generated/go/admin"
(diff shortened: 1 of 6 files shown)


services/controllerhost/controllerhost.go
```diff
@@ -173,7 +174,8 @@ func NewController(cfg configure.CommonAppConfig, sVice *common.Service, metadat
 
 	context.dstLock = lockMgr
 	context.m3Client = metrics.NewClient(instance.Service.GetMetricsReporter(), metrics.Controller)
-	context.mm = common.NewMetadataMgr(metadataClient, context.m3Client, context.log)
+	mClient := metadata.NewMetadataMetricsMgr(metadataClient, context.m3Client, context.log)
```
KiranRG: you should probably update the global (Mcp.mClient) to use the instance that emits metrics .. so that other sub-components (within controllerhost) that calls into metadata would automatically emit metrics. i see that retention manager instantiate another instance of metadata-metrics within its code -- we should probably remove that.

kirg (changes requested):

services/controllerhost/controllerhost.go
venkat1109: good idea, done !

services/controllerhost/controllerhost.go
```diff
@@ -173,7 +173,7 @@ func NewController(cfg configure.CommonAppConfig, sVice *common.Service, metadat
 
 	context.dstLock = lockMgr
 	context.m3Client = metrics.NewClient(instance.Service.GetMetricsReporter(), metrics.Controller)
-	context.mm = common.NewMetadataMgr(metadataClient, context.m3Client, context.log)
+	context.mm = NewMetadataMgr(metadataClient, context.m3Client, context.log)
```
venkat1109: done

kirg (approved):
LoGoToMe.

PR #48, opened 2017-01-30
RetMgr: handle empty extents correctly

KiranRG:
The change fixes the issue with retention-manager where it fails to move empty, sealed extents into "consumed" (and "deleted") states.

### DIFF
--- a/services/retentionmgr/metadataDep.go
+++ b/services/retentionmgr/metadataDep.go
@@ -394,12 +394,21 @@ func (t *metadataDepImpl) GetAckLevel(destID destinationID, extID extentID, cgID
 	// assert(resp.GetExtent().GetExtentUUID() == extID)
 	// assert(resp.GetExtent().GetConsumerGroupUUID() == cgID)
 
-	// check if the consumer-group has read to "sealed" point
-	if resp.GetExtent().GetStatus() != metadata.ConsumerGroupExtentStatus_OPEN {
-		ackLevel = store.ADDR_SEAL
-	} else {
+	switch resp.GetExtent().GetStatus() {
+
+	case metadata.ConsumerGroupExtentStatus_OPEN:
 		ackLevel = resp.GetExtent().GetAckLevelOffset()
-		// assert(ackLevel != cherami.ADDR_END
+
+	case metadata.ConsumerGroupExtentStatus_CONSUMED:
+		ackLevel = store.ADDR_SEAL
+
+	case metadata.ConsumerGroupExtentStatus_DELETED:
+		ackLevel = store.ADDR_BEGIN
+
+	default:
+		ackLevel = store.ADDR_BEGIN
+		log.WithField(`ConsumerGroupExtentStatus`, resp.GetExtent().GetStatus()).
+			Error("GetAckLevel: Unknown ConsumerGroupExtentStatus")
 	}
 
 	log.WithField(`ackLevel`, ackLevel).Debug("GetAckLevel done")
--- a/services/retentionmgr/metadataDep.go
+++ b/services/retentionmgr/metadataDep.go
@@ -397,12 +397,15 @@ func (t *metadataDepImpl) GetAckLevel(destID destinationID, extID extentID, cgID
 	switch resp.GetExtent().GetStatus() {
 
 	case metadata.ConsumerGroupExtentStatus_OPEN:
+		// return the ack-level from metadata
 		ackLevel = resp.GetExtent().GetAckLevelOffset()
 
 	case metadata.ConsumerGroupExtentStatus_CONSUMED:
+		// 'ADDR_SEAL' indicates to the caller that this CG has fully consumed the extent
 		ackLevel = store.ADDR_SEAL
 
 	case metadata.ConsumerGroupExtentStatus_DELETED:
+		// set to 'ADDR_BEGIN' if cg-extent is deleted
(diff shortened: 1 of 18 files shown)


services/retentionmgr/retention.go
```diff
@@ -696,21 +703,21 @@ func (t *RetentionManager) computeRetention(job *retentionJob, log bark.Logger)
...
 	//	   "consumed" the extent)
 	// B. or, the hard-retention has reached the end of the sealed extent,
 	// 	in which case we will force the extent to be "consumed"
-	// NB: retentionAddr == ADDR_BEGIN indicates there was an error, so we no-op
-	if job.retentionAddr != store.ADDR_BEGIN &&
-		((ext.status == shared.ExtentStatus_SEALED &&
-			minAckAddr == store.ADDR_SEAL &&
-			softRetentionConsumed) ||
-			hardRetentionConsumed) {
+	// NB: if there was an error querying either the hard/soft retention addresses,
```
datoug: can we simply to logic here by error out the retention directly when there's any metadata failure?

services/retentionmgr/metadataDep.go
```diff
@@ -394,12 +394,21 @@ func (t *metadataDepImpl) GetAckLevel(destID destinationID, extID extentID, cgID
...
-	if resp.GetExtent().GetStatus() != metadata.ConsumerGroupExtentStatus_OPEN {
-		ackLevel = store.ADDR_SEAL
-	} else {
+	switch resp.GetExtent().GetStatus() {
+
+	case metadata.ConsumerGroupExtentStatus_OPEN:
 		ackLevel = resp.GetExtent().GetAckLevelOffset()
-		// assert(ackLevel != cherami.ADDR_END
+
+	case metadata.ConsumerGroupExtentStatus_CONSUMED:
```
datoug: can we add some comments here regarding why the ack level is set to seal or begin?
also the change doesn't affect the behavior because you won't call this for deleted cg, right?

services/retentionmgr/retention.go
```diff
@@ -696,21 +703,21 @@ func (t *RetentionManager) computeRetention(job *retentionJob, log bark.Logger)
...
+	// {soft,hard}RetentionConsumed would be set to 'false'; if there was an error
+	// querying ack-addr, then allHaveConsumed will be false. therefore errors in
+	// either of the conditions would cause the extent to *not* be moved to the
+	// CONSUMED state, and would cause it to be retried on the next iteration.
+	if (ext.status == shared.ExtentStatus_SEALED) &&
+		((allHaveConsumed && softRetentionConsumed) || hardRetentionConsumed) {
```
venkat1109: this will default to applying `softRetention` if there are no consumer-groups. The safest thing to do is apply hardRetention. May be, you can special case if the extent has zero messages left, but that will be an optimization.

datoug: +1 Also I think this will break the multi_zone soft retention logic

services/retentionmgr/retention.go
KiranRG: I think it's not that simple. The thing is if there was an error with either querying the minAckAddr or the softRetentionAddr, but there no error querying the hardRetentionAddr, then we could potentially go ahead with marking the extent consumed, etc. I think this logic encapsulates all those cases, as expected.

services/retentionmgr/metadataDep.go
KiranRG: will do.

services/retentionmgr/retention.go
KiranRG: I agree .. I will special-case when there are no consumers and if this is a remote extent, to not move the extent to consumed.

Note: this diff does not affect the computation of the retention-address (it only affects when we move this to consumed). I shall address that as part of a separate change, for the zero-consumer case.

services/retentionmgr/retention.go
```diff
@@ -601,12 +604,15 @@ func (t *RetentionManager) computeRetention(job *retentionJob, log bark.Logger)
 		}
 	}
 
-	// If this is a multi_zone destination and local extent, disable soft retention
-	// The reason is if soft retention is very short, we may delete messages before remote zone has a chance to replicate the messages
-	// Long term solution should create a fake consumer for the remote zone
+	// If this is a multi_zone destination and local extent, disable soft retention -- and fall
+	// back to using only hard-retention. The reason is if soft retention is very short, we may
+	// delete messages before remote zone has a chance to replicate the messages Long term solution
```
datoug: Nit: let's add a space between 'messages Long '

services/retentionmgr/retention.go
```diff
@@ -649,22 +656,32 @@ func (t *RetentionManager) computeRetention(job *retentionJob, log bark.Logger)
...
 
 			minAckAddr = ackAddr
 		}
 	}
 
-	// if we were unable to find any consumer groups, set minAckAddr to ADDR_BEGIN
+	// if we were unable to find any consumer groups, set minAckAddr to
+	// ADDR_BEGIN, effectively disabling soft-retention. that said, hard
+	// retention could still be enforced.
 	if minAckAddr == store.ADDR_END {
```
datoug: I thought you also need to check len(job.consumers)==0?

KiranRG:
discarding this .. pull request; looks like a rebase messed up things, making tonnes of unrelated changes to be be shown as part of this diff.

PR #49, opened 2017-01-30
Miscellaneous improvements on outputhost

aravindvs:
1. Don't always refresh the CG during load. This was resulting
in one metadata cg extent call for every ReceiveMessageBatch call.
2. Convert spammy logs to metrics.
3. When the notify from controller is received, make sure to just
refresh the CG instead of loading completely.

### DIFF
--- a/common/metrics/defs.go
+++ b/common/metrics/defs.go
@@ -721,6 +721,8 @@ const (
 	OutputhostCGMessageSentNAck
 	// OutputhostCGMessagesThrottled records the count of messages throttled
 	OutputhostCGMessagesThrottled
+	// OutputhostCGAckMgrSeqNotFound is the gauge to track acks whose seq number is not found
+	OutputhostCGAckMgrSeqNotFound
 	// OutputhostCGMessageSentLatency is the latency to send a message
 	OutputhostCGMessageSentLatency
 	//OutputhostCGMessageCacheSize is the cashe size of consumer group message
@@ -745,8 +747,6 @@ const (
 	OutputhostCGAckMgrResetMsgError
 	// OutputhostCGSkippedMessages is the gauge to track skipped messages
 	OutputhostCGSkippedMessages
-	// OutputhostCGAckMgrSeqNotFound is the gauge to track acks whose seq number is not found
-	OutputhostCGAckMgrSeqNotFound
 	// OutputhostCGCreditsAccumulated is a gauge to record credits that are accumulated locally per consumer group
 	OutputhostCGCreditsAccumulated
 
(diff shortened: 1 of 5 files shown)


services/outputhost/cgcache.go
```diff
@@ -444,7 +444,23 @@ func (cgCache *consumerGroupCache) manageConsumerGroupCache() {
 	}
 }
 
-// refreshCgCache contacts metadata to get all the open extents and then loads them if needed
+// refreshCgCacheNoLock is a routine which is called without the extMutex held.
+// It takes the mutex and in turn refreshes the cg by contactingmetadata to
```
datoug: nit: space between contactingmetadata

datoug: kind of confused here, don't we hold the lock during the call? Why it's called XXXNoLock()?

common/metrics/defs.go
```diff
@@ -743,6 +745,10 @@ const (
 	OutputhostCGAckMgrResetMsgError
 	// OutputhostCGSkippedMessages is the gauge to track skipped messages
 	OutputhostCGSkippedMessages
+	// OutputhostCGAckMgrSeqNotFound is the gauge to track acks whose seq number is not found
+	OutputhostCGAckMgrSeqNotFound
```
venkat1109: this looks like it must be counter, any reason for this to be gauge ?

services/outputhost/cgcache.go
```diff
@@ -264,7 +264,7 @@ func newConsumerGroupCache(destPath string, cgDesc shared.ConsumerGroupDescripti
 		cfgMgr:                   h.cfgMgr,
 	}
 
-	cgCache.checkSingleCGVisibleCache = cache.NewLRUWithInitialCapacity(1, 1 << 6)
+	cgCache.checkSingleCGVisibleCache = cache.NewLRUWithInitialCapacity(1, 1<<6)
```
venkat1109: nit: 64 is more readable than 1<<6.. I know you didn't make this change, but isn't 64 too big for this cache ?

common/metrics/defs.go
aravindvs: Since it will be reset now and then Gauge made sense here.. we already have a CreditsReceived Counter to track the credits received here.. this is locally accumulated credits when it cannot be sent to the write pump at the moment..

services/outputhost/cgcache.go
aravindvs: It is big but since it is just the maxSize i think it is fine, since the initial size is still 1.. One more reason for making this 64 is i think @GuillaumeBailey wants to guarantee that this extent will definitely be in the cache to prevent the metadata call..

services/outputhost/cgcache.go
aravindvs: this routine, refreshCgCacheNoLock() gets called without the lock held..i.e, the caller doesn't hold the lock and it needs to be taken inside.. whereas the refreshCgCache() is the one where the caller already has the lock and a lock is not needed inside the routine..

common/metrics/defs.go
venkat1109: aravindvs@ - Question was about `OutputhostCGAckMgrSeqNotFound` just to be clear

aravindvs: Ah.. ok.. as we discussed in person, let me update this..

services/outputhost/cgcache.go
aravindvs: as we discussed in person, looks like the LRU cache is an overkill here and can be avoided.. we can address it as a separate patch..

common/metrics/defs.go
```diff
@@ -719,6 +721,8 @@ const (
 	OutputhostCGMessageSentNAck
 	// OutputhostCGMessagesThrottled records the count of messages throttled
 	OutputhostCGMessagesThrottled
+	// OutputhostCGAckMgrSeqNotFound is the gauge to track acks whose seq number is not found
```
venkat1109: fix comment - s/gauage/counter

PR #51, opened 2017-02-01
RetMgr: handle empty extents correctly

KiranRG:
This is a pull request corresponding to the earlier one (https://github.com/uber/cherami-server/pull/48) that was closed, because a rebase messed up the diff.

The changes here incorporate review feedback from the earlier request.


services/retentionmgr/retention.go
```diff
@@ -483,6 +483,9 @@ func (t *RetentionManager) computeRetention(job *retentionJob, log bark.Logger)
 		}
 	}
 
+	// isRemoteZoneExtent is set if the extent is a remote multi-zone extent
+	var isRemoteZoneExtent = dest.isMultiZone && !common.IsRemoteZoneExtent(ext.originZone, t.Options.LocalZone)
```
datoug: let's delete this one

services/retentionmgr/retention.go
```diff
@@ -649,22 +656,32 @@ func (t *RetentionManager) computeRetention(job *retentionJob, log bark.Logger)
...
 
 			minAckAddr = ackAddr
 		}
 	}
 
-	// if we were unable to find any consumer groups, set minAckAddr to ADDR_BEGIN
+	// if we were unable to find any consumer groups, set minAckAddr to
+	// ADDR_BEGIN, effectively disabling soft-retention. that said, hard
+	// retention could still be enforced.
 	if minAckAddr == store.ADDR_END {
```
datoug: I thought this needs to check len(job.consumers)==0?

KiranRG: True, but that check is redundant .. if there are no consumers, the minAckAddr would be ADDR_END (the default initialization value).

services/retentionmgr/retention.go
KiranRG: done.

PR #52, opened 2017-02-01
Light weight listExtents API impl for use in controller

venkat1109:
This patch provides an implementation of APIs previously added to the interface as part of this https://github.com/uber/cherami-thrift/pull/7 - the goal is to expose light weight listExtents api that fetches only the interesting columns for controller. This patch also does a glide up to pick up the thrift changes. Other packages are also updated as part of the glide up. Manually inspected the commits for gorocksdb/cherami-client-go/gocql and confirmed they are safe.

### DIFF
--- a/clients/metadata/metadata_cassandra.go
+++ b/clients/metadata/metadata_cassandra.go
@@ -2727,7 +2727,7 @@ func (s *CassandraMetadataService) listExtentsStatsHelper(request *shared.ListEx
 	return query.Iter(), nil
 }
 
-// ListDestinationExtens lists all the extents mapped to a given destination
+// ListDestinationExtents lists all the extents mapped to a given destination
 func (s *CassandraMetadataService) ListDestinationExtents(ctx thrift.Context, request *m.ListDestinationExtentsRequest) (*m.ListDestinationExtentsResult_, error) {
 
 	if len(request.GetDestinationUUID()) == 0 {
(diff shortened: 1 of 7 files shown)


clients/metadata/metadata_cassandra.go
```diff
@@ -3703,6 +3798,81 @@ func (s *CassandraMetadataService) readConsumerGroupExtentsHelper(request *m.Rea
 	return query.Iter(), nil
 }
 
+// ReadConsumerGroupExtentsLite returns the list all extents mapped to
+// the given consumer group. This API only returns a few interesting
+// columns for each extent in the result. For detailed info about
+// extents, see ReadConsumerGroupExtents
+func (s *CassandraMetadataService) ReadConsumerGroupExtentsLite(ctx thrift.Context, request *m.ReadConsumerGroupExtentsLiteRequest) (*m.ReadConsumerGroupExtentsLiteResult_, error) {
+
+	if request.GetMaxResults() < 1 {
```
KiranRG: nit: in 'ListDestinationExtents', your corresponding check is "<=0"; might want to make it consistent .. ;-)

clients/metadata/metadata_cassandra.go
```diff
@@ -3703,6 +3798,81 @@ func (s *CassandraMetadataService) readConsumerGroupExtentsHelper(request *m.Rea
...
+		filterLocally = true
+	}
+
+	sql := sqlCGGetAllExtentsLite
+	if !filterLocally && request.IsSetStatus() {
+		sql += fmt.Sprintf(" AND %s=%d", columnStatus, request.GetStatus())
+	}
+
+	qry := s.session.Query(sql).Consistency(s.lowConsLevel).Bind(request.GetConsumerGroupUUID())
+	if request.GetMaxResults() > 0 {
```
KiranRG: isn't this check redundant? since we enforce 'GetMaxResults' to be at least '1' above?

clients/metadata/metadata_cassandra.go
venkat1109: sure, will do

clients/metadata/metadata_cassandra.go
venkat1109: it may be.. I will remove it if it is

PR #53, opened 2017-02-01
RetMgr should potentially run on non-primary controller

KiranRG:
Retention Manager uses 'FindHostForKey' to determine the controller host that it should run on. It uses a different key from what controller uses (to determine the primary controller), but it seems like the ret-mgr seems to always pick the primary controller as the host to run on.

I found that the root-cause is that 'FindHostForKey' (in common/rpm.go) ignore the "key" in picking the host. This change addresses that by using a hash of the given key to pick the host.

### DIFF
--- a/common/rpm.go
+++ b/common/rpm.go
@@ -247,34 +247,25 @@ func (rpm *ringpopMonitorImpl) FindHostForAddr(service string, addr string) (*Ho
 // FindHostForKey finds and returns the host responsible for handling the given key
 func (rpm *ringpopMonitorImpl) FindHostForKey(service string, key string) (*HostInfo, error) {
 
-	// get list of hosts that for the given service
-	members, err := rpm.GetHosts(service)
-	if err != nil || members == nil || len(members) == 0 {
-		return nil, ErrUnknownService
-	}
-
 	// this function should be consistent in picking and returning the
-	// same node for a given key on a specific 'set' of hosts; so we sort
-	// all the hosts by their ip-addresses and use a hash of the key to
-	// pick the host from that list.
-	var ipAddrs []string
-	hostInfoMap := make(map[string]*HostInfo, len(members))
-	for _, member := range members {
-		ipAddrs = append(ipAddrs, member.Addr)
-		hostInfoMap[member.Addr] = member
-	}
-
-	// choose an ipAddr using hash of the given key
-	sort.Strings(ipAddrs) // sort ip-addresses
+	// same node for a given key on a specific 'set' of hosts. the list
+	// of hosts returned by rpm.GetHosts is guaranteed to be sorted by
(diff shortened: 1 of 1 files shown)


common/rpm.go
```diff
@@ -245,27 +246,35 @@ func (rpm *ringpopMonitorImpl) FindHostForAddr(service string, addr string) (*Ho
...
-	// We need to make sure we always return the same node
-	// So sort members by IP and return the lowest.
+	// this function should be consistent in picking and returning the
+	// same node for a given key on a specific 'set' of hosts; so we sort
+	// all the hosts by their ip-addresses and use a hash of the key to
+	// pick the host from that list.
 	var ipAddrs []string
-	tempMap := make(map[string]int, len(members))
-	for k, member := range members {
+	hostInfoMap := make(map[string]*HostInfo, len(members))
```
venkat1109: I think we can avoid the temp map creation and sorting for every call. 
How about we keep the membershipInfo.asList() sorted, when we create it (see rpm.refresh()) ?
Then all this method has to do is - 

members := GetHosts()
return members[hash%len(members)].Addr

common/rpm.go
```diff
@@ -245,27 +246,26 @@ func (rpm *ringpopMonitorImpl) FindHostForAddr(service string, addr string) (*Ho
...
+	// hash the key and use it to pick the host from the list.
+
+	// compute FNV-1a hash of the key
+	fnv1a := fnv.New32a()
+	fnv1a.Write([]byte(key))
+	hash := int(fnv1a.Sum32())
+
+	// get list of hosts that for the given service
 	members, err := rpm.GetHosts(service)
 	if err != nil || members == nil || len(members) == 0 {
```
venkat1109: GetHosts() already does these checks and returns the correct err code.
This function can just return err, if err != nil

common/rpm.go
```diff
@@ -245,27 +246,26 @@ func (rpm *ringpopMonitorImpl) FindHostForAddr(service string, addr string) (*Ho
...
 		return nil, ErrUnknownService
```
venkat1109: leave the err code as is, as mentioned above.

common/rpm.go
```diff
@@ -245,27 +246,26 @@ func (rpm *ringpopMonitorImpl) FindHostForAddr(service string, addr string) (*Ho
...
-		// maintain the reverse mapping to have the index
-		tempMap[member.Addr] = k
-	}
-	// sort the ip addresses
-	sort.Strings(ipAddrs)
-	// get the key of the lowest
-	lowestIndex := tempMap[ipAddrs[0]]
-	return members[lowestIndex], nil
+	// pick the host corresponding to the hash
+	return members[hash%len(members)], nil
```
venkat1109: returning a copy of hostInfo is safer, since rpm assumes hostInfo to be immutable.

PR #58, opened 2017-02-03
Admin command to check if sealed extents are sealed on replicas

KiranRG:
This adds an admin command that iterates through extents and checks to ensure that if the extent is sealed in metadata, that it is sealed on all of the replicas, as well.

### DIFF
--- a/clients/storehost/client.go
+++ b/clients/storehost/client.go
@@ -83,3 +83,11 @@ func (s *StoreClientImpl) SealExtent(req *store.SealExtentRequest) error {
 
 	return s.client.SealExtent(ctx, req)
 }
+
+// PurgeMessages seals an extent on the specified store
+func (s *StoreClientImpl) PurgeMessages(req *store.PurgeMessagesRequest) (*store.PurgeMessagesResult_, error) {
+	ctx, cancel := tcthrift.NewContext(2 * time.Second)
+	defer cancel()
+
+	return s.client.PurgeMessages(ctx, req)
+}
--- a/clients/storehost/client.go
+++ b/clients/storehost/client.go
@@ -76,7 +76,7 @@ func (s *StoreClientImpl) GetAddressFromTimestamp(req *store.GetAddressFromTimes
 	return s.client.GetAddressFromTimestamp(ctx, req)
 }
 
-// SealExtent queries store for the address corresponding to the given timestamp
+// SealExtent seals an extent on the specified store
 func (s *StoreClientImpl) SealExtent(req *store.SealExtentRequest) error {
 	ctx, cancel := tcthrift.NewContext(2 * time.Second)
 	defer cancel()
--- a/clients/storehost/client.go
+++ b/clients/storehost/client.go
@@ -67,3 +67,19 @@ func (s *StoreClientImpl) ReadMessages(req *store.ReadMessagesRequest) (*store.R
 
 	return s.client.ReadMessages(ctx, req)
 }
+
+// GetAddressFromTimestamp queries store for the address corresponding to the given timestamp
+func (s *StoreClientImpl) GetAddressFromTimestamp(req *store.GetAddressFromTimestampRequest) (*store.GetAddressFromTimestampResult_, error) {
+	ctx, cancel := tcthrift.NewContext(2 * time.Second)
+	defer cancel()
+
+	return s.client.GetAddressFromTimestamp(ctx, req)
+}
+
+// SealExtent queries store for the address corresponding to the given timestamp
+func (s *StoreClientImpl) SealExtent(req *store.SealExtentRequest) error {
+	ctx, cancel := tcthrift.NewContext(2 * time.Second)
+	defer cancel()
+
+	return s.client.SealExtent(ctx, req)
+}
--- a/tools/common/lib.go
+++ b/tools/common/lib.go
@@ -1153,13 +1153,18 @@ func SealConsistencyCheck(c *cli.Context, mClient mcli.Client) {
 		Limit:  common.Int64Ptr(DefaultPageSize),
 	}
 
+iterate_listdestinations_pages:
 	for {
 		if veryVerbose {
 			fmt.Printf("querying metadata: ListDestinations(prefix=\"%s\")\n", prefix)
 		}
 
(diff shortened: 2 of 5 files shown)


clients/storehost/client.go
```diff
@@ -67,3 +67,19 @@ func (s *StoreClientImpl) ReadMessages(req *store.ReadMessagesRequest) (*store.R
...
+
+// GetAddressFromTimestamp queries store for the address corresponding to the given timestamp
+func (s *StoreClientImpl) GetAddressFromTimestamp(req *store.GetAddressFromTimestampRequest) (*store.GetAddressFromTimestampResult_, error) {
+	ctx, cancel := tcthrift.NewContext(2 * time.Second)
+	defer cancel()
+
+	return s.client.GetAddressFromTimestamp(ctx, req)
+}
+
+// SealExtent queries store for the address corresponding to the given timestamp
```
venkat1109: fix comment

tools/common/lib.go
```diff
@@ -1079,3 +1080,148 @@ func getIfSetString(c *cli.Context, p string, setCount *int) (r *string) {
...
+	return client, nil
+}
+
+func (t *storeClientCache) close() {
+	for _, client := range t.cache {
+		client.Close()
+	}
+}
+
+// SealConsistencyCheck implement for show msg command line
```
venkat1109: fix comment

tools/common/lib.go
```diff
@@ -1079,3 +1080,148 @@ func getIfSetString(c *cli.Context, p string, setCount *int) (r *string) {
...
+func SealConsistencyCheck(c *cli.Context, mClient mcli.Client) {
+
+	/*
```
venkat1109: remove dead code

tools/common/lib.go
```diff
@@ -1079,3 +1080,148 @@ func getIfSetString(c *cli.Context, p string, setCount *int) (r *string) {
...
+}
+
+func (t *storeClientCache) get(storeUUID string) (*storehost.StoreClientImpl, error) {
+
+	client, ok := t.cache[storeUUID]
+
+	if !ok {
+		hostAddr, err := t.mClient.UUIDToHostAddr(storeUUID)
+		if err != nil {
+			hostAddr = storeUUID + UnknownUUID
```
venkat1109: what does this line do ?

tools/common/lib.go
```diff
@@ -1079,3 +1080,148 @@ func getIfSetString(c *cli.Context, p string, setCount *int) (r *string) {
...
+}
+
+func (t *storeClientCache) close() {
+	for _, client := range t.cache {
+		client.Close()
+	}
+}
+
+// SealConsistencyCheck implement for show msg command line
+func SealConsistencyCheck(c *cli.Context, mClient mcli.Client) {
```
venkat1109: how about moving the scc into its own file ?

clients/storehost/client.go
```diff
@@ -67,3 +67,19 @@ func (s *StoreClientImpl) ReadMessages(req *store.ReadMessagesRequest) (*store.R
...
+func (s *StoreClientImpl) SealExtent(req *store.SealExtentRequest) error {
```
venkat1109: didn't see this API used, remove if not needed.

tools/common/lib.go
```diff
@@ -1079,3 +1080,188 @@ func getIfSetString(c *cli.Context, p string, setCount *int) (r *string) {
...
+								// now seal the extent
+								if seal {
+
+									fmt.Printf("sealing extent on replica: ", destUUID, extentUUID, storeUUID)
+
+									req := store.NewSealExtentRequest()
+									req.ExtentUUID = common.StringPtr(string(extentUUID))
+									req.SequenceNumber = nil // seal at 'unspecified' seqnum
+
+									// query storage to find address of the message with the given timestamp
```
datoug: update comment

tools/common/lib.go
```diff
@@ -1079,3 +1080,188 @@ func getIfSetString(c *cli.Context, p string, setCount *int) (r *string) {
...
+								fmt.Printf("dest=%v extent=%v store=%v: ", destUUID, extentUUID, storeUUID)
+
+								if extentNotFound {
+									fmt.Printf("not sealed (extent missing)\n", destUUID, extentUUID, storeUUID)
+								} else {
+									fmt.Printf("not sealed\n", destUUID, extentUUID, storeUUID)
+								}
+
+								// now seal the extent
+								if seal {
```
datoug: so we want to send seal request even if the extent is not found?

tools/common/lib.go
```diff
@@ -1079,3 +1080,188 @@ func getIfSetString(c *cli.Context, p string, setCount *int) (r *string) {
...
+
+		for _, desc := range resp.GetDestinations() {
+
+			destUUID := desc.GetDestinationUUID()
+
+			// fmt.Printf("destUUID=%v\n", destUUID)
+
+			listExtentsStats := &shared.ListExtentsStatsRequest{
+				DestinationUUID: common.StringPtr(string(destUUID)),
+				Limit:           common.Int64Ptr(DefaultPageSize),
```
datoug: you can filtered out remote extent by setting localExtentsOnly=true

clients/storehost/client.go
KiranRG: done.

clients/storehost/client.go
KiranRG: it is used now.

tools/common/lib.go
KiranRG: i am not 100% sure .. but that seems to be the pattern we have used throughout the code. my sense is it is to ensure we print it out the error with the UUID, etc.

tools/common/lib.go
KiranRG: thanks for that; done.

tools/common/lib.go
KiranRG: yes; there are scenarios where a store does not even know about an extent .. but it has already been sealed. sending a seal request to a store for such an extent, creates that extent and seals it.

tools/common/lib.go
KiranRG: done.

PR #59, opened 2017-02-03
add health check in dockerfile

datoug:


docker/server/Dockerfile
```diff
@@ -29,7 +29,12 @@ RUN cd $CHERAMI_HOME; make bins
...
 
+COPY ./healthcheck.sh $CHERAMI_HOME/healthcheck.sh
+RUN chmod a+x $CHERAMI_HOME/healthcheck.sh
+
 WORKDIR $CHERAMI_HOME
 
 # Remember to mount host directories (config and data), and have CHERAMI_CONFIG_DIR properly point to config.
 CMD ["./cherami-server", "start", "all"]
+
+HEALTHCHECK --interval=1s --timeout=1s --retries=120 CMD ./healthcheck.sh
```
thuningxu: Why do this on server container?  The "server" container can be started without frontend inside (e.g. docker run ... "./cherami-server start input output"), then this check will fail.

docker/standalone/start.sh
```diff
@@ -24,7 +24,7 @@
 pushd /
 ./docker-entrypoint.sh cassandra
 popd
-sleep 20
+sleep 40
```
thuningxu: How about someway to health check cassandra (e.g. via `cqlsh -e 'desc system'`) instead of fixed sleep?

thuningxu:
There are couple of problems:
1. I don't think the docker build is successful (see below logs). 
2. I also think you should try to do a docker build only on a master branch.
3. Looks like you want to build a docker from local files. The Dockerfile does not support that because it retrieves master from github.

```
Step 16 : COPY ./start.sh $CHERAMI_HOME/start.sh
lstat start.sh: no such file or directory
after_success.7
0.02s$ docker tag $REPO:$COMMIT $REPO:$TAG
Error response from daemon: no such id: ubercherami/cherami-server-standalone:829b4874
after_success.8
0.02s$ docker tag $REPO:$COMMIT $REPO:travis-$TRAVIS_BUILD_NUMBER
Error response from daemon: no such id: ubercherami/cherami-server-standalone:829b4874
after_success.9
0.03s$ docker push $REPO
The push refers to a repository [docker.io/ubercherami/cherami-server-standalone]
An image does not exist locally with the tag: ubercherami/cherami-server-standalone

```

PR #60, opened 2017-02-03
randomly select a replicator host for replication jobs to even out load

datoug:


aravindvs (approved):

common/rpm.go
```diff
@@ -275,6 +277,16 @@ func (rpm *ringpopMonitorImpl) FindHostForKey(service string, key string) (*Host
 	return &host, nil
 }
 
+// FindRandomHostForKey finds and returns a random host responsible for handling the given key
+func (rpm *ringpopMonitorImpl) FindRandomHostForKey(service string) (*HostInfo, error) {
+	members, err := rpm.GetHosts(service)
+	if err != nil || members == nil || len(members) == 0 {
```
venkat1109: GetHosts() already does all of these checks for you and returns the correct error code. Just check for err != nil here

common/rpm.go
```diff
@@ -65,6 +65,8 @@ type (
 		FindHostForAddr(service string, addr string) (*HostInfo, error)
 		// FindHostForKey finds and returns the host responsible for handling the given (service, key)
 		FindHostForKey(service string, key string) (*HostInfo, error)
+		// FindRandomHostForKey finds and returns a random host responsible for handling the given service
+		FindRandomHostForKey(service string) (*HostInfo, error)
```
venkat1109: s/FindRandomHostForKey/FindRandomHost - there is no key arg to this function.

PR #61, opened 2017-02-04
Build rocksdb once from source

thuningxu:
Instead of using a prebuilt package, build rocksdb once, and dynamically link it in tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,13 +29,13 @@ install:
   - go get golang.org/x/tools/cmd/cover
   - ccm create test -v 2.2.8 -n 1 -s
   - sudo ln -sf /home/travis/.local/bin/cqlsh /usr/local/bin/cqlsh
-  - wget https://github.com/uber/cherami-server/releases/download/rocksdb-4.11.2-trusty/librocksdb.so.4.11.2
-  - ln -s librocksdb.so.4.11.2 librocksdb.so.4.11
-  - ln -s librocksdb.so.4.11.2 librocksdb.so.4
-  - ln -s librocksdb.so.4.11.2 librocksdb.so
+  - glide install && touch vendor/glide.updated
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -29,13 +29,13 @@ install:
   - go get golang.org/x/tools/cmd/cover
   - ccm create test -v 2.2.8 -n 1 -s
   - sudo ln -sf /home/travis/.local/bin/cqlsh /usr/local/bin/cqlsh
-  - wget https://github.com/uber/cherami-server/releases/download/rocksdb-4.11.2-trusty/librocksdb.so.4.11.2
-  - ln -s librocksdb.so.4.11.2 librocksdb.so.4.11
-  - ln -s librocksdb.so.4.11.2 librocksdb.so.4
-  - ln -s librocksdb.so.4.11.2 librocksdb.so
+  - glide install && touch vendor/glide.updated
+  - pushd vendor/github.com/cockroachdb/c-rocksdb/internal
+  - travis_wait 20 make shared_lib
```
venkat1109: isn't this going to make builds flaky ?

thuningxu: Why? 20 minutes should be good enough for building rocksdb (typically < 600 seconds).

venkat1109: ah.. numbers without units always baffle me.. nvm

aravindvs: If pinning rocksdb to a specific version is the issue, let's revert this and make sure our glide.yaml for the appropriate rocksdb libs are pinned..

Looks like building rocksdb now is the bottleneck for running tests.

Whenever we decide to use the latest rocksdb, we can update this to pick the latest and have a pre-built version ready as well.

venkat1109: +1.. make shared_lib now takes ~450-500s every time.. that's 10 mins for every build..

PR #63, opened 2017-02-06
Improve queue depth performance with aggressive caching

GuillaumeBailey:
Queue depth is getting slower and slower. The available evidence is that the store extent stats metadata lookups are a major preventable cause of slowness. In the existing design, we do not cache these extent stats between queue depth runs, because controller was meant the scale vertically, and there was a concern about excessive memory use.

We will soon switch to horizontal scaling for queue depth calculations, which means that we can afford to have a permanent cache for extent statistics, much of which is static. For now, this acts as a solution for latency-induced performance problems in queue depth. 

* Move from small, ephemeral MRU cache to larger, persistent LRU cache
* Reduce the size of cached data by removing unnecessary fields and indirection
* Deprecate unavailable backlog, which is only for timer destinations (deprecated) and never had necessary data from store host

### DIFF
--- a/services/controllerhost/queueDepth.go
+++ b/services/controllerhost/queueDepth.go
@@ -77,10 +77,12 @@ type QueueDepthCacheJSONFields struct {
 	BacklogDLQ         int64               `json:"backlog_dlq"`
 }
 
+type extentCoverMap map[extentID]struct{}
+
 //var testQueueDepth bool
 var queueDepthCacheLk sync.Mutex
 var queueDepthCache = make(map[string]QueueDepthCacheEntry) // key is CG UUID
-var queueDepthCacheLimit = 1 << 20
+var queueDepthCacheLimit = 1 << 19
 
 type gaftKey struct {
 	extentUUID string
@@ -93,16 +95,32 @@ type queueDepthCalculator struct {
 	*extentStateMonitor
 	ll                      bark.Logger
 	destinationExtentsCache replicaStatsMRUCache // MRU cache of extent replica stats
-	dlqExtentsCache         replicaStatsMRUCache // MRU cache of dlq extent replica stats
+	dstCoverMap             extentCoverMap
+	dlqCoverMap             extentCoverMap
 }
 
 func newQueueDepthCalculator(m *extentStateMonitor) *queueDepthCalculator {
 	return &queueDepthCalculator{
 		extentStateMonitor:      m,
-		destinationExtentsCache: *newReplicaStatsMRUCache(m.context.mm, m.context.log.WithField(`module`, `ReplicaStatsMRUCache`)),
-		dlqExtentsCache:         *newReplicaStatsMRUCache(m.context.mm, m.context.log.WithField(`module`, `ReplicaStatsMRUCacheDLQ`)),
-		ll:                      m.context.log.WithField(`module`, `queueDepth`),
+		destinationExtentsCache: *newReplicaStatsMRUCache(m.context.mm, m.context.log.WithField(`module`, `ReplicaStatsMRUCache`), queueDepthCacheLimit),
+		ll:          m.context.log.WithField(`module`, `queueDepth`),
+		dstCoverMap: make(extentCoverMap),
+		dlqCoverMap: make(extentCoverMap),
 	}
+}
+
+// Makes a copy of the cover map that is safe to mutate
+func (qdc *queueDepthCalculator) getCoverMap(dlq bool) (r extentCoverMap) {
+	orig := qdc.dstCoverMap
+	if dlq {
+		orig = qdc.dlqCoverMap
+	}
+
+	r = make(extentCoverMap, len(orig))
+	for ext := range orig {
+		r[ext] = struct{}{}
+	}
+	return r
 }
 
 func (qdc *queueDepthCalculator) processDestination(dstDesc *shared.DestinationDescription, dlqCache bool) {
@@ -115,12 +133,13 @@ func (qdc *queueDepthCalculator) processDestination(dstDesc *shared.DestinationD
 		return
 	}
 
-	cache := &qdc.destinationExtentsCache
+	coverMap := &qdc.dstCoverMap
 	if dlqCache {
-		cache = &qdc.dlqExtentsCache
+		coverMap = &qdc.dlqCoverMap
 	}
 
-	cache.clear()
+	// Clear the coverMap
+	*coverMap = make(extentCoverMap)
 
 	filter := []shared.ExtentStatus{shared.ExtentStatus_OPEN, shared.ExtentStatus_SEALED} // CONSUMED should be ignorable for queue depth
 	extents, err = context.mm.ListDestinationExtentsByStatus(dstDesc.GetDestinationUUID(), filter)
@@ -132,15 +151,13 @@ func (qdc *queueDepthCalculator) processDestination(dstDesc *shared.DestinationD
 
 	for _, ext := range extents {
 		extID := extentID(ext.GetExtentUUID())
-		cache.putSingleCGVisibility(extID, ext.GetConsumerGroupVisibility())
-		for _, store := range ext.GetStoreUUIDs() {
-			cache.put(extID, storeID(store), nil) // Add a placeholder so that a store extent may be found if no consumer group extent exists for it
-		}
+		// This updates the consumer group visibility for DLQ merge and purge, and also our store UUIDs for re-replication
+		qdc.destinationExtentsCache.putPlaceholders(extID, ext.GetConsumerGroupVisibility(), ext.GetStoreUUIDs())
+		(*coverMap)[extentID(ext.GetExtentUUID())] = struct{}{}
 	}
 
-	if !dlqCache { // Don't process queue depth for DLQ CGs
+	if !dlqCache { // Don't process queue depth for DLQ CGs, just populate the DLQ coverMap and extent cache
 		qdc.processConsumerGroups(dstDesc)
-		cache.clear()
 	}
 }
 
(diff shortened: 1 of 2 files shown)


services/controllerhost/replicaStatsMRUCache.go
```diff
@@ -127,82 +132,134 @@ func (c *replicaStatsMRUCache) get(extent extentID, store storeID) (val *shared.
...
 		return nil
-	} else if val != nil {
-		c.put(extent, store, val)
+	} else if extStats != nil {
+		val = makeCompactExtentStats(extStats)
+		// In the typical case, all store extents had planned sealing, meaning that their available sequences are the same
+		// Map the value pointer to the same compactExtentStats in this typical case, to reduce memory usage by nearly 66%
+	dedup:
+		for dedupStore, val2 = range psmap.m {
+			if store != dedupStore && val2 != nil && val2.seqAvailable == val.seqAvailable {
```
venkat1109: don't follow.. How can you de-dup compactExtentStats across different extents ?

services/controllerhost/replicaStatsMRUCache.go
```diff
@@ -77,47 +78,51 @@ func (c *replicaStatsMRUCache) get(extent extentID, store storeID) (val *shared.
...
-	if len(store) == 0 {
-		store = foundStore //pick a random store if none specified
+	if len(store) == 0 && val != nil {
+		goto ttlCheck // return any found store if none specified, if applicable
 	}
 
+	// Cache miss, perform lookup and try to put
 	if len(store) == 0 {
-		return // This happens when a new extent comes online and there are various race conditions in Cassandra
+		store = foundStore //pick a random store if none specified
```
venkat1109: don't follow - isn't this going to reset the passed arg `store` ? Why do this ? And what's the guarantee foundStore is != nil ?

services/controllerhost/replicaStatsMRUCache.go
```diff
@@ -127,82 +132,134 @@ func (c *replicaStatsMRUCache) get(extent extentID, store storeID) (val *shared.
...
 	return
-}
 
-func (c *replicaStatsMRUCache) putSingleCGVisibility(extent extentID, singleCGVisibility string) {
-	if len(extent) == 0 || len(singleCGVisibility) == 0 {
-		return
+ttlCheck:
+
+	// Open extents should only be valid for a single scan; All types of sealed extent should have indefinite retention
+	if val != nil && val.status == shared.ExtentStatus_OPEN && time.Since(val.cacheTime) > IntervalBtwnScans {
```
venkat1109: if the TTL is intervalBtwnScans, is the purpose of this cache to avoid additional lookups when there are multiple consumer-groups for the same destination ?

services/controllerhost/replicaStatsMRUCache.go
```diff
@@ -127,82 +132,134 @@ func (c *replicaStatsMRUCache) get(extent extentID, store storeID) (val *shared.
...
-func (c *replicaStatsMRUCache) put(extent extentID, store storeID, val *shared.ExtentStats) {
-	var ok bool
-	var psmap *perStoreReplicaStatsMap
-
-	if len(extent) == 0 || len(store) == 0 {
-		c.log.
-			WithField(common.TagExt, string(extent)). /* Don't call common.Fmt with an empty string ! */
-			WithField(common.TagStor, string(store)).
-			Warn(`Queue depth skipped empty ID`)
+func (c *replicaStatsMRUCache) putPlaceholders(extent extentID, singleCGVisibility string, storeIDs []string) {
```
venkat1109: How will  this one be used ?

services/controllerhost/queueDepth.go
```diff
@@ -93,16 +95,32 @@ type queueDepthCalculator struct {
 	*extentStateMonitor
 	ll                      bark.Logger
 	destinationExtentsCache replicaStatsMRUCache // MRU cache of extent replica stats
-	dlqExtentsCache         replicaStatsMRUCache // MRU cache of dlq extent replica stats
+	dstCoverMap             extentCoverMap
```
venkat1109: can you add some comments on what these coverMaps are ? Why are they being reset inside processDestination() ?

services/controllerhost/queueDepth.go
```diff
@@ -325,66 +340,53 @@ func (qdc *queueDepthCalculator) calculateAndReportMetrics(dstDesc *shared.Desti
...
+				`runningTotalAvail`: ba + BacklogAvailable,
 			}).Info(`Queue Depth Tabulation`)
 		}
 
 		BacklogAvailable += ba
 		BacklogUnavailable += bu
 	}
 
-	coverMap = qdc.destinationExtentsCache.getExtentCoverMap()
+	coverMap = qdc.getCoverMap(false)
```
venkat1109: why make a copy here ?

services/controllerhost/replicaStatsMRUCache.go
```diff
@@ -77,47 +78,51 @@ func (c *replicaStatsMRUCache) get(extent extentID, store storeID) (val *shared.
...
-		}
-
-		// Get any one store's replica stats; see if it is sealed
-	oneStore:
-		for foundStore, val = range psmap.m {
-			if val != nil {
-				break oneStore
-			}
-		}
+	intrfc = c.cache.Get(string(extent))
```
venkat1109: Is this cache used across destinations ? If so, same extentID across 2 destinations will stomp on each other.

GuillaumeBailey:
> (1) When you say, the calculation is getting slower, is it the entire process i.e, processDestination->processConsumerGroups->processConsumerGroupExtents->calculateAndReportMetrics or is it just the calculateAndReportMetrics taking a long time? Do we need to add latency around these APIs to measure which takes a longer time before optimizing code?

The whole process takes a long time, but what I could tell from the logs, is that just running through the store extents was taking multiple seconds per consumer group. That's the inner-loop, and there's no reason that it couldn't take microseconds, at least for the sealed extents.

(2) Looking at the cherami-metadata graph, it looks like store extents stats calls p99 seems to hover around 450 to 500ms; Note: It is not saved as part of the dashboard yet (operation:metadatareadstoreextentreplicastats). So may be this might not be a bottle neck?

From the logs, it strongly appears to be the bottleneck. The logs suggest that the average is around 100ms, which is making many consumer groups take >10 seconds to process.

(3) Can we use the *Lite version of metadata APIs here or does it not have enough info on them?

The lite version cuts out all the info that I would need.

services/controllerhost/queueDepth.go
GuillaumeBailey: The existing code deletes entries from the cover map. An explicit copy must be made so that multiple consumer groups under a destination can have a mutable copy of the extent map.

services/controllerhost/replicaStatsMRUCache.go
GuillaumeBailey: added a comment

`// putPlaceholders is used by processDestination to pre-populate/update some data from the destination extent stats on every pass
// * The single consumer group visibility, which changes even on sealed extents due to DLQ merge/purge operations
// * The list of owning store IDs, which (in the future) will change with re-replication, archival activities, etc.`

services/controllerhost/queueDepth.go
```diff
@@ -93,16 +95,32 @@ type queueDepthCalculator struct {
...
+	dlqCoverMap             extentCoverMap
 }
 
 func newQueueDepthCalculator(m *extentStateMonitor) *queueDepthCalculator {
 	return &queueDepthCalculator{
 		extentStateMonitor:      m,
-		destinationExtentsCache: *newReplicaStatsMRUCache(m.context.mm, m.context.log.WithField(`module`, `ReplicaStatsMRUCache`)),
-		dlqExtentsCache:         *newReplicaStatsMRUCache(m.context.mm, m.context.log.WithField(`module`, `ReplicaStatsMRUCacheDLQ`)),
-		ll:                      m.context.log.WithField(`module`, `queueDepth`),
+		destinationExtentsCache: *newReplicaStatsMRUCache(m.context.mm, m.context.log.WithField(`module`, `ReplicaStatsMRUCache`), queueDepthCacheLimit),
```
GuillaumeBailey: I bet this will make the diff incomprehensible, so I'll make a 2nd diff to do rename/restructuring.

services/controllerhost/replicaStatsMRUCache.go
GuillaumeBailey: The original MRU implementation already achieved that goal.

The purpose is to achieve what you said, but also to avoid lookups for the sealed extents across multiple scans, which form the majority of the work. 

Lookups on the opened extents are unavoidable, but will become much better with auto-scale (e.g. only 1 or 2 open extents for slowly moving destinations).

services/controllerhost/replicaStatsMRUCache.go
GuillaumeBailey: It's the same extent, but we have multiple copies of the store extent because the particular store that we read from is important in the unplanned seal case. We would otherwise get 'off by a few' phantom backlog.

This optimization reduces memory use for the normal case that all replicas for a sealed extent agree on the available sequence.

services/controllerhost/replicaStatsMRUCache.go
GuillaumeBailey: added some comments to explain.

The particular replica we get doesn't matter unless we had an unplanned seal.

services/controllerhost/replicaStatsMRUCache.go
GuillaumeBailey: This isn't a problem. It is true that we have merged extents that exist in two destinations, but they are deleted in one of them. putPlaceholders ensures that each destination gets the proper view of the extent (i.e. different singleCGVisibility), even though the underlying replica stats doesn't change. 

So merge doesn't cause a problem here. We don't calculate queue depth for consumer groups of the DLQ, either. Nor do we calculate queue depth for the DLQ destination itself.

services/controllerhost/queueDepth.go
GuillaumeBailey: done

GuillaumeBailey:
image

This is a graphical view of the log analysis that I did earlier. It shows the progress of queue depth calculation for a single consumer group. Note that while most of the underlying metadata operations are completing quickly, the cumulative time here is dominated by the outlier lookups.

PR #64, opened 2017-02-07
Output admin commands

aravindvs:
This patch adds a couple of admin commands on outputhost to inspect the
in-memory cg and ack manager states.

(1) The first command is to list all loaded consumer groups on a
specific outputhost
(2) The second command is to read the cg state on an outputhost given
the cgUUID. The cg state includes all in-memory info about this cg and
also the AckManager for all the extents within this CG. This will be
helpful in finding "how many holes are present in the ack manager since
the last ack level".

### DIFF
--- a/cmd/tools/admin/main.go
+++ b/cmd/tools/admin/main.go
@@ -619,7 +619,7 @@ func main() {
 						cli.StringFlag{
 							Name:  "cg_uuid, cg",
 							Value: "",
-							Usage: "The consumergroup UUID whose in-memory state to dump",
+							Usage: "The UUID of the consumer group whose state will be dumped",
 						},
 					},
 					Action: func(c *cli.Context) {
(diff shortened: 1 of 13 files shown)


glide.yaml
```diff
@@ -34,6 +34,7 @@ import:
   - common/websocket
   - stream
 - package: github.com/uber/cherami-thrift
+  version: output_admin
```
aravindvs: TODO: will fix this once the corresponding diff on the cherami-thrift repo is landed.

cmd/tools/admin/main.go
```diff
@@ -629,6 +607,50 @@ func main() {
...
+			Subcommands: []cli.Command{
+				{
+					Name:    "cgstate",
+					Aliases: []string{"cgs"},
+					Usage:   "outputhost cgstate <hostport> [options]",
+					Flags: []cli.Flag{
+						cli.StringFlag{
+							Name:  "cg_uuid, cg",
+							Value: "",
+							Usage: "The consumergroup UUID whose in-memory state to dump",
```
GuillaumeBailey: The UUID of the consumer group whose state will be dumped

PR #67, opened 2017-02-10
Add dstPath+cgName as parameters for queue depth info

GuillaumeBailey:


services/frontendhost/frontend.go
```diff
@@ -1301,12 +1301,25 @@ func (h *Frontend) GetQueueDepthInfo(ctx thrift.Context, queueRequest *c.GetQueu
 	if _, err = h.prolog(ctx, queueRequest); err != nil {
 		return
 	}
+	var cgDesc *c.ConsumerGroupDescription
 
-	cgUUID := queueRequest.GetKey()
+	cgUUID := queueRequest.GetConsumerGroupName()
```
venkat1109: doesn't look right, you want the name or uuid ?

services/frontendhost/frontend.go
```diff
@@ -1301,12 +1301,25 @@ func (h *Frontend) GetQueueDepthInfo(ctx thrift.Context, queueRequest *c.GetQueu
...
+			DestinationPath:   common.StringPtr(queueRequest.GetDestinationPath()),
+		}
+		cgDesc, err = h.ReadConsumerGroup(ctx, rcgReq)
+		if err != nil {
+			return
+		}
+
+		cgUUID = cgDesc.GetConsumerGroupUUID()
+	} else {
+		if queueRequest.GetDestinationPath() != `` {
```
venkat1109: Suggest restructuring this code to make it clear to the reader... How about the following:

var cgUUID string 


if len(request.GetDestinationPath()) > 0 {
 // when dstPth is specified, the name param must be a cg name
   if  isUUID(request.GetConsumerGroupName()) {
      // error out
   }
   cgUUID = lookupUUID(name)
} else {
   // when dstPth is empty, name is *overloaded* as uuid
   cgUUID = request.GetConsumerGroupName()
} 


if !isUUID(cgUUID) {
   // return error
}

services/frontendhost/frontend.go
GuillaumeBailey: This is similar to other APIs, where ConsumerGroupName can be the name or (maybe) the UUID. This naming is consistent with the 95% use case of specifying destination path and name, and consistent with the 5% case of overloading it with the UUID, similar to other Frontend APIs

services/frontendhost/frontend.go
GuillaumeBailey: done.

tools/common/lib.go
```diff
@@ -546,20 +546,30 @@ func ReadDlq(c *cli.Context, mClient mcli.Client) {
...
 	if len(c.Args()) < 1 {
 		ExitIfError(errors.New(strNotEnoughArgs))
 	}
 
-	cgUUID := c.Args()[0]
+	switch len(c.Args()) {
+	case 1:
+		cg = c.Args()[0]
+	default:
+		cg = c.Args()[1]
```
venkat1109: Can you also add a len() check here and exit gracefully, if the number of args < expected ?

PR #69, opened 2017-02-13
Add size based seal notification on inputhost

aravindvs:
This patch simply uses the existing API to notify controller when an
extent reaches it's size limit. Add a unit test for the same.

Also fix a subtle race in setting of the lastSuccessSeqNo which got
exposed as part of the new unit test.

Right now, the size of the extent is just based on the max allowed seq number,
assuming an average message size of 1K. So we can get a max size of 20GB
assuming our max rollover seq num is 20mill.

In the future, may be, we can make this a dynamically configurable item per destination.

### DIFF
--- a/services/inputhost/exthost.go
+++ b/services/inputhost/exthost.go
@@ -107,6 +107,9 @@ type (
 		lastExtLoadReportedTime int64 // unix nanos when the last extent metrics were reported
 
 		minimumAllowedMessageDelaySeconds int32 // min delay on messages
+
+		maxSizeMB  float32 // max size for this extent; TODO: Make this dynamically configurable?
+		currSizeMB float32 // current size of this extent
 	}
 
 	// Holds a particular extent for use by multiple publisher connections.
@@ -158,6 +161,13 @@ const (
 
 	// extLoadReportingInterval is the interval destination extent load is reported to controller
 	extLoadReportingInterval = 2 * time.Second
+
+	// extMaxSize is the maximum size for this extent after which we notify the controller to seal
+	// based on an average of 1K message sizes, we allow the extent to grow upto 20GB by default
+	extMaxSizeMB float32 = 20000
+
+	// bytesPerMB is used to maintain the size in bytes
+	bytesPerMB float32 = 1024 * 1024
 )
 
 var (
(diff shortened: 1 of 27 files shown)


services/inputhost/exthost.go
```diff
@@ -158,6 +161,13 @@ const (
 
 	// extLoadReportingInterval is the interval destination extent load is reported to controller
 	extLoadReportingInterval = 2 * time.Second
+
+	// extMaxSize is the maximum size for this extent after which we notify the controller to seal
+	// based on an average of 1K message sizes, we allow the extent to grow upto 20GB by default
+	extMaxSizeMB float32 = 20000
+
+	// bytesPerMB is used to maintain the size in bytes
+	bytesPerMB float32 = 1024 * 1024
```
venkat1109: extMaxSizeMB, bytesPerMB - name/value consistency - Let's either use MB (1000) or MiB (1024).

services/inputhost/exthost.go
```diff
@@ -568,11 +584,16 @@ func (conn *extHost) sendMessage(pr *inPutMessage, extSendTimer *common.Timer, w
 		return
 	}
 
-	// If we reach the max sequence number, notify the extent controller but
-	// keep the pumps open
+	// just get the size of the actual data here
+	// ignoring other headers for the size computation
+	// Note: size of will not work here
+	conn.currSizeMB += float32(len(pr.putMsg.Data)) / bytesPerMB
```
venkat1109: how about tracking the currSize and maxSize as bytes instead of MiBs ? That way, they can just be int64s and you also don't need the additional math conversion for every message.

services/inputhost/exthost.go
```diff
@@ -158,6 +161,13 @@ const (
 
 	// extLoadReportingInterval is the interval destination extent load is reported to controller
 	extLoadReportingInterval = 2 * time.Second
+
+	// extMaxSize is the maximum size for this extent after which we notify the controller to seal
+	// based on an average of 1K message sizes, we allow the extent to grow upto 20GB by default
+	extMaxSizeMB float32 = 20000
```
venkat1109: this const isn't used anywhere.

services/inputhost/exthost.go
```diff
@@ -179,25 +189,29 @@ var (
...
+		streamClosedChannel:    make(chan struct{}),
+		numReplicas:            numReplicas,
+		shutdownWG:             shutdownWG,
+		forceUnloadCh:          make(chan struct{}),
+		limitsEnabled:          limitsEnabled,
+		maxSequenceNumber:      common.GetRandInt64(int64(extentRolloverSeqnumMin), int64(extentRolloverSeqnumMax)),
+		// by default, max size is now based on the max allowed sequence number
+		// and an average message size of 1k;
+		// So in general, it will be max of 20GB((or 20000MB)), assuming max seq num is 20 million (or 20000MB)
+		maxSizeMB:               float32(extentRolloverSeqnumMax*1024) / bytesPerMB,
```
venkat1109: 20GB is too big IMO. Let's say a single extent cannot exceed more than 1% of total disk capacity. For 1.2T HDD, this number comes to ~12G. Suggest going with a fixed number <=1% and lets not make the size a function of the maxCount.

services/inputhost/exthost.go
```diff
@@ -628,7 +649,7 @@ func (conn *extHost) aggregateAndSendReplies(numReplicas int) {
 					}
 				}
 				// mark this seqNo as the last success seqno
-				conn.lastSuccessSeqNo = resCh.seqNo
+				atomic.StoreInt64(&conn.lastSuccessSeqNo, resCh.seqNo)
```
venkat1109: wow !

services/inputhost/exthost.go
venkat1109: how about also making this a randomized range ?

aravindvs: i actually had it 20G because we have 2T HDD (most of our prod machines are as such) and so 1% of this ~20G. But I agree we can make this randomized based on the sealSeqNumber..

services/inputhost/exthost.go
aravindvs: ACK.. will change it..

services/inputhost/exthost.go
```diff
@@ -304,8 +319,10 @@ func (conn *extHost) close() {
 	conn.lk.Unlock() // no longer need the lock
 
 	conn.logger.WithFields(bark.Fields{
-		`sentSeqNo`: conn.seqNo,
-		`ackSeqNo`:  conn.lastSuccessSeqNo,
+		`sentSeqNo`:      conn.seqNo,
+		`ackSeqNo`:       conn.lastSuccessSeqNo,
+		`currentSizeMiB`: conn.getSizeInMiB(conn.currSizeBytes),
```
venkat1109: just print the currSizeBytes and maxSizeBytes as such, to avoid any rounding and/or precision errors while conversion.  this can be valuable info for debugging.

PR #72, opened 2017-02-15
replicator metadata reconcilation should handle deleted extents in remote

datoug:
Here, remote means the origin zone. 
1. if remote is in deleted state, and local doesn't have the extent, don't create the extent locally
2. if remote is in deleted state, local is in open/sealed, try sealing it in store
3. if remote doesn't have the extent, local is in open/sealed, seal it after 24 hrs

### DIFF
--- a/services/replicator/metadataReconciler.go
+++ b/services/replicator/metadataReconciler.go
@@ -474,7 +474,7 @@ func (r *metadataReconciler) reconcileDestExtent(destUUID string, localExtents m
 				continue
 			}
 		} else {
-			if remoteExtentStatus == shared.ExtentStatus_SEALED && localExtentStatus == shared.ExtentStatus_OPEN {
+			if (remoteExtentStatus == shared.ExtentStatus_SEALED || remoteExtentStatus == shared.ExtentStatus_CONSUMED) && localExtentStatus == shared.ExtentStatus_OPEN {
 				r.sealExtentInMetadata(destUUID, remoteExtentUUID)
 			}
 			if remoteExtentStatus == shared.ExtentStatus_DELETED {
--- a/services/replicator/metadataReconciler.go
+++ b/services/replicator/metadataReconciler.go
@@ -21,6 +21,7 @@
 package replicator
 
 import (
+	"errors"
 	"strings"
 	"sync/atomic"
 	"time"
@@ -32,6 +33,7 @@ import (
 	"github.com/uber/cherami-server/common/metrics"
 	"github.com/uber/cherami-thrift/.generated/go/metadata"
 	"github.com/uber/cherami-thrift/.generated/go/shared"
+	"github.com/uber/cherami-thrift/.generated/go/store"
 )
 
 type (
@@ -44,8 +46,9 @@ type (
 
 	// metadataReconciler is an implementation of MetadataReconciler.
 	metadataReconciler struct {
-		replicator *Replicator
-		localZone  string
+		replicator            *Replicator
+		localZone             string
+		suspectMissingExtents map[string]missingExtentInfo
 
 		mClient  metadata.TChanMetadataService
 		logger   bark.Logger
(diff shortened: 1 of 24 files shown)


services/replicator/metadataReconciler.go
```diff
@@ -451,37 +470,170 @@ func (r *metadataReconciler) reconcileDestExtent(destUUID string, localExtents m
...
+	for suspectExtent, suspectExtentInfo := range r.suspectMissingExtents {
+		if suspectExtentInfo.destUUID != destUUID {
+			continue
+		}
+		if _, ok := remoteMissingExtents[suspectExtent]; !ok {
+			// re-appeared in remote, remove it from suspect list
+			delete(r.suspectMissingExtents, suspectExtent)
+		} else {
+			if time.Since(suspectExtentInfo.missingSince) > extentMissingDurationThreshold {
+				r.handleExtentDeletedOrMissingInRemote(suspectExtentInfo.destUUID, suspectExtent, localExtents[suspectExtent])
```
venkat1109: (1) if there is an ongoing replication job, shouldn't the status of the job be marked as DONE here ?
(2) Make sure localExtents[id] returns ok, if not, delete the extent from the suspect list (i.e. suspect missing from local & remote) and not re-add it back in line 517

services/replicator/metadataReconciler.go
```diff
@@ -451,37 +470,170 @@ func (r *metadataReconciler) reconcileDestExtent(destUUID string, localExtents m
...
+
+	for suspectExtent, suspectExtentInfo := range r.suspectMissingExtents {
+		if suspectExtentInfo.destUUID != destUUID {
+			continue
+		}
+		if _, ok := remoteMissingExtents[suspectExtent]; !ok {
+			// re-appeared in remote, remove it from suspect list
+			delete(r.suspectMissingExtents, suspectExtent)
+		} else {
+			if time.Since(suspectExtentInfo.missingSince) > extentMissingDurationThreshold {
```
venkat1109: you also need to remove the extent from the suspect list after handleExtentDeletedOrMiss..(), if that method returns success. But be careful not to re-add in line 517.

services/replicator/metadataReconciler.go
```diff
@@ -423,10 +434,18 @@ func (r *metadataReconciler) reconcileDestExtent(destUUID string, localExtents m
...
-				common.TagZoneName: common.FmtZoneName(remoteZone),
+				common.TagDst:          common.FmtDst(destUUID),
+				common.TagExt:          common.FmtExt(remoteExtentUUID),
+				common.TagZoneName:     common.FmtZoneName(remoteZone),
+				common.TagExtentStatus: common.FmtExtentStatus(remoteExtentStatus),
 			}).Warn(`Found missing extent from remote!`)
+
+			// if the extent is already in consumed/deleted state on remote side, don't bother to create the extent locally because all data is gone on remote side already
+			if remoteExtentStatus == shared.ExtentStatus_CONSUMED || remoteExtentStatus == shared.ExtentStatus_DELETED {
+				r.m3Client.UpdateGauge(metrics.ReplicatorReconcileScope, metrics.ReplicatorReconcileDestExtentRemoteDeletedLocalMissing, 1)
```
venkat1109: suggest separate metric for consumed

services/replicator/metadataReconciler.go
```diff
@@ -451,37 +470,170 @@ func (r *metadataReconciler) reconcileDestExtent(destUUID string, localExtents m
...
+	// now add new suspect if there's any
+	for missingExtent, _ := range remoteMissingExtents {
+		if _, ok := r.suspectMissingExtents[missingExtent]; !ok {
+			r.suspectMissingExtents[missingExtent] = missingExtentInfo{
+				destUUID:     destUUID,
+				missingSince: time.Now(),
 			}
 		}
 	}
+
```
venkat1109: also, would be cool to emit a metric on len(suspectMissingExtents) at the end of each reconciliation.

services/replicator/metadataReconciler.go
```diff
@@ -451,37 +470,170 @@ func (r *metadataReconciler) reconcileDestExtent(destUUID string, localExtents m
...
-				r.logger.WithFields(bark.Fields{
-					common.TagDst: common.FmtDst(destUUID),
-					common.TagExt: common.FmtExt(remoteExtentUUID),
-				}).Info(`Extent SEALED`)
+	// note we're not 100% sure if the extent is really gone even if cassandra doesn't return that extent because of
+	// the eventual consistency nature of cassandra
+	// so we maintain a timestamp to track how long the extent has been missing
+	// we act on it only after it has been missing for a certain period of time
+
+	remoteMissingExtents := make(map[string]bool)
```
venkat1109: map[string]struct{}

services/replicator/metadataReconciler.go
```diff
@@ -451,37 +474,179 @@ func (r *metadataReconciler) reconcileDestExtent(destUUID string, localExtents m
...
+				common.TagErr:  err,
+				common.TagStor: common.FmtStor(store),
+			}).Error(`Failed to seal extent on store`)
+			errorOccured = true
+			continue
+		}
+	}
+
+	if errorOccured {
+		return errors.New(`seal extent on store failed for at least one replica`)
```
KiranRG: our mechanism that detects inconsistency on an extent being sealed on metadata vs a replica skips xDC-extents. and if we encounter an error here, how can we ensure this does get sealed eventually? (as in is there another fallback mechanism?)

services/replicator/metadataReconciler.go
```diff
@@ -451,37 +474,179 @@ func (r *metadataReconciler) reconcileDestExtent(destUUID string, localExtents m
...
 			}
 		} else {
-			if remoteExtentStatus != shared.ExtentStatus_OPEN && localExtentStatus == shared.ExtentStatus_OPEN {
-				r.logger.WithFields(bark.Fields{
-					common.TagDst:      common.FmtDst(destUUID),
-					common.TagExt:      common.FmtExt(remoteExtentUUID),
-					common.TagZoneName: common.FmtZoneName(remoteZone),
-				}).Info(`Inconsistent extent status`)
-				r.m3Client.UpdateGauge(metrics.ReplicatorReconcileScope, metrics.ReplicatorReconcileDestExtentInconsistentStatus, 1)
+			if remoteExtentStatus == shared.ExtentStatus_SEALED && localExtentStatus == shared.ExtentStatus_OPEN {
```
datoug: added 'consumed' scenario. As discussed offline, since handleExtentDeletedOrMissingInRemote is a special/corner case, I feel like using a switch is not very suitable here.

services/replicator/metadataReconciler.go
datoug: the next reconciliation run will try again (and will keeping trying until it's sealed). I'm planning to add more metrics/stats in next diff to monitor this.

PR #73, opened 2017-02-17
Replicator fix deadlock

datoug:
the way replicator uses GetClientFactory causes a deadlock because we try to get it before starting service. Simple fix is to remove the variable.

### DIFF
--- a/services/replicator/metadataReconciler.go
+++ b/services/replicator/metadataReconciler.go
@@ -621,7 +621,7 @@ func (r *metadataReconciler) sealExtentInStore(destUUID string, extentUUID strin
 
 	var errorOccured bool
 	for _, store := range readExtentResult.GetExtentStats().GetExtent().GetStoreUUIDs() {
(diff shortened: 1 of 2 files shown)


services/replicator/metadataReconciler.go
```diff
@@ -71,7 +71,7 @@ const (
 	runInterval                    = time.Duration(10 * time.Minute)
 	metadataListRequestPageSize    = 50
 	storeCallTimeout               = 10 * time.Second
-	extentMissingDurationThreshold = time.Duration(24 * time.Hour)
+	extentMissingDurationThreshold = time.Duration(1 * time.Hour)
```
venkat1109: Consider making this a tunable, if you have dynamic config already hooked in. 
If not, lets do this as a followup.

datoug: Good point. I have it but it seems not working. Will look at that first then switch over.

PR #75, opened 2017-02-17
fix a replicator reconilation bug

datoug:
We used to get all local extents and compare with remote, which includes extents that originated from the local zone. The reconciliation really should only care about the extents that are originated from the remote zone, so adding a filtering before the reconciliation.

Also fix the ReplicatorReconcileDestExtentRemoteDeletedLocalNot metrics issue


services/replicator/metadataReconciler.go
```diff
@@ -352,7 +352,15 @@ func (r *metadataReconciler) reconcileDestExtentMetadata() error {
 					continue
 				}
 
-				if err = r.reconcileDestExtent(dest.GetDestinationUUID(), localExtents, remoteExtents, zoneConfig.GetZone()); err != nil {
+				// only need to reconcile for extents that are originated from the remote zone
+				filteredLocalExtents := make(map[string]shared.ExtentStatus)
```
venkat1109: how about moving this inside getAllDestExtentInCurrentZone() ?

venkat1109: Also does the same problem exist in getAllDestExtentInRemoteZone() ?

datoug: 1. can we move this inside getAllDestExtentInCurrentZone?
No because the filtering is done per zone. Let's say we have zone a,b,c. getAllDestExtentInCurrentZone return extents a1,a2,b1,b2,c1,c2. If current zone is a, remote zone is b, then we only want to have b1,b2 as filteredLocalExtents

2. does the same problem exist in getAllDestExtentInRemoteZone
getAllDestExtentInRemoteZone we filtered on local extent only: https://github.com/uber/cherami-server/blob/master/services/replicator/metadataReconciler.go#L377

PR #78, opened 2017-02-21
Fix code coverage: make integ tests count

venkat1109:
Currently, our integration tests don't count towards code coverage. There are a couple of problems:

* go test needs a special argument if the test lives outside of the package
* correct merging of cover files matters

This patch addresses both of the above above problems and updates the makefiles. 

After the change, running the integration test discovered an unexercised code path referencing a non-existent symbol. That's also fixed (clients/controller/client.go).


venkat1109:
Linking the relevant golang issue here for clarification:

https://github.com/golang/go/issues/6909

PR #82, opened 2017-02-28
Tighten ack aggregation on input host

GuillaumeBailey:
Line 632 of the original is a latent bug; Status_OK = 0, which means that the zero value for ack (given when okCh = false) has Status_OK. I don't believe that anyone closes the appendMsgAck channel, but it is better to fix this now.

Adjusted so that we set stat exactly twice, once to make it default to failed, and then at most once to update it to the actual status.

One additional failure mode is covered (desynchronization of receipts).
One uncovered failure mode remains: receiving more than one ack from a single replica.

Clarified some comments and LOG-only code.

### DIFF
--- a/services/inputhost/exthost.go
+++ b/services/inputhost/exthost.go
@@ -625,33 +625,54 @@ func (conn *extHost) aggregateAndSendReplies(numReplicas int) {
 
 				// Note: even if this value is negative, it is ok because we should timeout immediately
 				perMsgTimer.Reset(msgAckTimeout - elapsed)
+
+				// TODO: tighten this further, by ensuring that we received exactly one ack from each replica, i.e. not N from 1 replica
+			aggregation:
 				for i := 0; i < numReplicas; i++ {
+					stat = cherami.Status_FAILED
 					select {
 					case ack, okCh := <-resCh.appendMsgAck:
-						if !okCh || ack.GetStatus() != cherami.Status_OK {
-							stat = ack.GetStatus()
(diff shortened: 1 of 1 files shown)


services/inputhost/exthost.go
```diff
@@ -625,33 +625,54 @@ func (conn *extHost) aggregateAndSendReplies(numReplicas int) {
 
 				// Note: even if this value is negative, it is ok because we should timeout immediately
 				perMsgTimer.Reset(msgAckTimeout - elapsed)
+
+				// TODO: tighten this further, by ensuring that we received exactly one ack from each replica, i.e. not N from 1 replica
```
aravindvs: We cannot receive more than one ACK from a replica because in replicaConnection.go we ensure that is not the case.

services/inputhost/exthost.go
```diff
@@ -625,33 +625,54 @@ func (conn *extHost) aggregateAndSendReplies(numReplicas int) {
...
+			aggregation:
 				for i := 0; i < numReplicas; i++ {
+					stat = cherami.Status_FAILED
 					select {
 					case ack, okCh := <-resCh.appendMsgAck:
-						if !okCh || ack.GetStatus() != cherami.Status_OK {
-							stat = ack.GetStatus()
-							// error means we shutdown this extent and seal it
-							go conn.close()
+						if !okCh { // channel closed, ack has all zero values
```
aravindvs: good catch.. but as you said this won't happen since nobody closes this channel..

PR #83, opened 2017-03-02
Fix extent replica stats reporting

KiranRG:
The way the current mechanism ('queueMontor') is designed, where in a periodic manner (once a minute) we take snapshot of all the active extents and write out extent-replica-stats, we may not update the stats on extents that are opened for short-operations (such as PurgeMessages, SealExtent, etc). I have refactored the mechanism completely in this change .. where we use callbacks from extent-manager to report stats on all extents that were opened .. by any operation that updates ('mutates') the extents.


services/storehost/extStats.go
```diff
@@ -0,0 +1,434 @@
...
+				// EndTime
+				// WriteTime
+			}
+
+			req := &metadata.UpdateStoreExtentReplicaStatsRequest{
+				ExtentUUID:   common.StringPtr(extentID.String()),
+				ReplicaStats: []*shared.ExtentReplicaStats{extReplStats},
+			}
+
+			err := t.mClient.UpdateStoreExtentReplicaStats(nil, req)
```
venkat1109: I believe this call can still overwrite TTLd entries (for an extent in DELETED state in metadata).One option would be read before update..

services/storehost/extStats.go
```diff
@@ -0,0 +1,434 @@
...
+	t.RLock()
+	defer t.RUnlock()
+
+	ctx, ok := t.extents[string(id)]
+
+	if ok && ctx.ext == ext {
+
+		// remove ref, if it drops to zero, then send out one last report
+		if atomic.AddInt64(&ctx.ref, -1) == 0 {
+			t.trySendReport(id, ext, ctx)
```
venkat1109: not sure if this is the right thing to do.. GAFT() calls open/close extents all the time..

services/storehost/extStats.go
```diff
@@ -0,0 +1,434 @@
...
+				t.RUnlock()
+
+				if atomic.LoadInt64(&ctx.ref) > 0 {
+
+					ext := ctx.ext
+
+					// trySendReport will be accessing 'ext' members,
+					// some of which require holding 'ext' shared.
+					ext.RLock()
+					t.trySendReport(uuid.UUID(id), ext, ctx)
```
KiranRG: worst case, that will cause a duplicate report to be sent ..

services/storehost/extStats.go
KiranRG: we filter in calls that only 'mutate' the extent .. so we will not send out reports in response to GAFT calls.

services/storehost/extStats.go
KiranRG: as discussed offline, there's two options here:
1. we should not expect to do this on a deleted extent -- because we update the store-extent-replica-stats only in response to an operation on store that 'mutates' the extent; and we should not expect to see any such calls onto the store after the extent has been marked for deletion.
2. if we do plan to do the check to see if it is deleted, i think it would be best to do it within the metadata-client api itself; and it's probably best to address that in a separate change.

services/storehost/extStats.go
```diff
@@ -0,0 +1,434 @@
...
+
+// ExtStatsReporterPause pauses the reporting (intended for tests)
+func ExtStatsReporterPause() {
+
+	swapped := atomic.CompareAndSwapInt32(&extStatsReporterPaused, 0, 1)
+	common.GetDefaultLogger().WithField(`already-paused`, !swapped).Info("extStatsReporter: paused")
+}
+
+// ExtStatsReporterUnpause unpauses the reporting (intended for tests)
+func ExtStatsReporterUnpause() {
```
KiranRG: done. i left it as 'unpause' because that's the name used by 'queueMonitor" (that this supplants) and in the integration_test that calls into this. nevertheless, i made this 'resume' and changed the call within integration_test too.

services/storehost/extMgr.go
```diff
@@ -39,6 +39,32 @@ import (
...
+		// ExtentOpen defines the callback function called when an extent is referenced;
+		// this is called immediately after InitCallback and for every open that comes
+		// in while the extent is 'active';
+		// the callback is called with extent-lock held shared.
+		ExtentOpen(id uuid.UUID, ext *extentContext, intent OpenIntent)
+
+		// ExtentClose defines the callback function called when an extent is dereferenced;
+		// the extent could still be 'active' due to other references;
+		// called with extent-lock held shared.
+		ExtentClose(id uuid.UUID, ext *extentContext, intent OpenIntent) (done bool)
```
venkat1109: Can you add a comment about the return value (bool ?)

services/storehost/extStats.go
```diff
@@ -0,0 +1,455 @@
...
+		case <-ticker.C:
+
+			// list of extents that can be deleted
+			deleteExtents := make([]string, 8)
+
+			// get lock shared, while iterating through map
+			t.extents.RLock()
+
+			// pause between extents to spread out the calls
+			pause := time.Duration(interval.Nanoseconds() / int64(1+len(t.extents.xmap)))
```
venkat1109: how about rate limiting calls to metadata on the reporterPump and having this pump block on reportC, instead of attempting to spread the load here ? If I understand right, you spread out load is to avoid browning out metadata.. so, say, reportingInteval=1minute and the maxUpdateReplicaStatsPerSecAllowed=10, with the current code, if the number of extents is 600, things will even out nicely. But if the number of extents is 1000, you will exceed the metadata limit.

services/storehost/extStats.go
```diff
@@ -0,0 +1,455 @@
...
+	}
+
+	r := t.reportPool.Get().(*report)
+
+	r.ctx = ctx
+	r.extentID = extentID
+	r.timestamp = time.Now().UnixNano()
+
+	// get a snapshot of various extent stats
+	r.firstSeqNum, r.lastSeqNum = ext.getBeginSeqNum(), ext.getLastSeqNum()
```
venkat1109: we also need the timestamp of the last message in the extent (or some way to derive it from the info you publish without making additional calls to store)..

services/storehost/extStats.go
```diff
@@ -0,0 +1,455 @@
...
+// ExtentInit is the callback from extent-manager when a new extent is initialized
+func (t *ExtStatsReporter) ExtentInit(id uuid.UUID, ext *extentContext) {
+	// no-op
+	return
+}
+
+// ExtentOpen is the callback from extent-manager when an extent is opened/referenced
+func (t *ExtStatsReporter) ExtentOpen(id uuid.UUID, ext *extentContext, intent OpenIntent) {
+
+	if intent != OpenIntentAppendStream &&
```
venkat1109: how about moving this into a helper ?

services/storehost/extStats.go
```diff
@@ -0,0 +1,455 @@
...
+		// 	common.TagExt:  extentID,      // #perfdisable
+		// 	`first-seqnum`: r.firstSeqNum, // #perfdisable
+		// 	`last-seqnum`:  r.lastSeqNum,  // #perfdisable
+		// }).Info("extStatsReporter: report dropped") // #perfdisable
+
+		return false // not sent
+	}
+}
+
+func (t *ExtStatsReporter) schedulerPump() {
```
venkat1109: this method is a great place to emit a metric that gives the count of mutated extents during one time slot. 
Also, would be cool to cool to emit other latency/count metrics from extStats

services/storehost/extMgr.go
```diff
@@ -400,34 +515,19 @@ func (xMgr *ExtentManager) IsExtentOpenedForReplication(extentID string) bool {
...
-func (xMgr *ExtentManager) getAllExtentsSeqNumSnapshot() map[string]*seqNumSnapshot {
-
-	xMgr.RLock()
-	defer xMgr.RUnlock()
+	// add to list of callback objects
+	xMgr.callbacks = append(xMgr.callbacks, callbacks)
+}
 
-	snapshot := make(map[string]*seqNumSnapshot, len(xMgr.extents))
+func (xMgr *ExtentManager) CallbackDone(ext *extentContext) {
```
venkat1109: where is this referenced ?

services/storehost/extMgr.go
```diff
@@ -105,6 +134,9 @@ type (
 		//    ref is 0, the 'extentLock' needs to be acquired and the ref re-checked
 		//    before initiating clean-up/tear-down
 
+		// count of close/cleanup callbacks that have reference on the structure
+		cleanupRef uint32
```
venkat1109: hmm.. don't really like this another refCounting for callbacks.. lets talk about this today.

services/storehost/extStats.go
KiranRG: yes, i have another diff in the works, that publishes start/end timestamps and addresses -- in order to fill out all the other fields in the extent-stats-report.

services/storehost/extStats.go
KiranRG: this is just a mechanism to report on all extents every minute .. with a best effort to spread that out. but yes, this does not rate-limit the calls to metadata per-se. if it does do so .. then we need other mechanisms to ensure we don't build up backlog internally, etc.

lets discuss about this .. should we perhaps tackle this in a separate change?

services/storehost/extStats.go
KiranRG: will do.

services/storehost/extStats.go
KiranRG: we already have metrics that show the number of open extents and number of active read/write streams, and other calls. so i think these metrics would be a little redundant ..

services/storehost/extMgr.go
KiranRG: this is a mechanism used by the callback-handlers, if they do not want the extent-context to be torn down. this is not used by the 'extStats' reporter -- which is the only component currently using this; but i plan to add others pretty soon.

services/storehost/extMgr.go
KiranRG: see comment above .. on mechanism to have callbacks to hold onto the extent context. this function is called when they are 'done' (and want to dereference) the extent-context.

services/storehost/extMgr.go
KiranRG: will do.

PR #85, opened 2017-03-06
Attempt to fix flaky Storetest

thuningxu:


services/storehost/storehost_test.go
```diff
@@ -38,6 +39,20 @@ import (
 	"github.com/uber/cherami-thrift/.generated/go/store"
 )
 
+func timedWgWait(wg *sync.WaitGroup, timeout time.Duration) bool {
```
KiranRG: "common.AwaitGroup" already does exactly this .. if you'd like to use it.

services/storehost/storehost_test.go
```diff
@@ -69,8 +84,8 @@ func (s *StoreHostSuite) TestStoreHostTimerQueueWriteWithRead() {
 		numExtents = 256                    // number of concurrent extents
 		minMessages, maxMessages = 0, 1000  // messages to send on an extent
 		minMsgSize, maxMsgSize = 512, 2048  // size of message
-		minMsgDelay, maxMsgDelay = 5, 300   // delay to use on messages (in seconds)
-		minRandWait, maxRandWait = 0, 10000 // wait between messages published (in milliseconds)
+		minMsgDelay, maxMsgDelay = 5, 30   // delay to use on messages (in seconds)
+		minRandWait, maxRandWait = 0, 1000 // wait between messages published (in milliseconds)
```
KiranRG: these values were intended to apply to a super-long test that needs to be initiated manually .. (using the "-long" flag).

services/storehost/storehost_test.go
```diff
@@ -85,10 +100,9 @@ func (s *StoreHostSuite) TestStoreHostTimerQueueWriteWithRead() {
 		var statsLatency = make([]*stats, numExtents)
 		var statsDrift = make([]*stats, numExtents)
 
+		wgAll.Add(numExtents)
+		var readerTimeout int32
 		for x := 0; x < numExtents; x++ {
-
-			wgAll.Add(1)
-
```
KiranRG: this is okay .. but just thought i'd let you know, this is a pattern i tend to use .. where the wait-group add immediately precedes the go-func call; and the first thing the go-func does is a 'defer waitgroup-done'.

thuningxu: Honestly, I am also not sure why this makes the tests stable.

thuningxu:
@KiranRG Note that I also applied same logic to `TestStoreHostTimerQueueWriteThenRead()`.

PR #87, opened 2017-03-08
add metrics to count max consecutive failures for replication job

datoug:


services/storehost/replicationJobRunner.go
```diff
@@ -224,6 +234,19 @@ func (runner *replicationJobRunner) run() {
 		}).Info(`replication for extent started`)
 	}
 
+	var maxConsecutiveFailures int
+	for existingFailedJob, failedTimes := range runner.failedJobs {
+		if _, ok := currentFailedJobs[existingFailedJob]; !ok {
+			delete(runner.failedJobs, existingFailedJob)
+		} else {
+			runner.failedJobs[existingFailedJob] = failedTimes + 1
+			if runner.failedJobs[existingFailedJob] > maxConsecutiveFailures {
```
venkat1109: you can avoid another map lookup by using your local var.

services/storehost/replicationJobRunner.go
```diff
@@ -224,6 +234,19 @@ func (runner *replicationJobRunner) run() {
 		}).Info(`replication for extent started`)
 	}
 
+	var maxConsecutiveFailures int
+	for existingFailedJob, failedTimes := range runner.failedJobs {
+		if _, ok := currentFailedJobs[existingFailedJob]; !ok {
```
venkat1109: If the job is missing in the currentFailedJobs, is there any guarantee that the job actually was started / finished from the time we last checked this map ?

services/storehost/replicationJobRunner.go
```diff
@@ -224,6 +234,19 @@ func (runner *replicationJobRunner) run() {
...
+			delete(runner.failedJobs, existingFailedJob)
+		} else {
+			runner.failedJobs[existingFailedJob] = failedTimes + 1
+			if runner.failedJobs[existingFailedJob] > maxConsecutiveFailures {
+				maxConsecutiveFailures = runner.failedJobs[existingFailedJob]
+			}
+		}
+	}
+	runner.m3Client.UpdateGauge(metrics.ReplicateExtentScope, metrics.StorageReplicationJobMaxConsecutiveFailures, int64(maxConsecutiveFailures))
```
venkat1109: would be nice to also log the job id that's failing consecutively..

venkat1109: also, would it makes sense to emit a simple counter on total number of job failures ? Or do we already have it ?

services/storehost/replicationJobRunner.go
```diff
@@ -224,9 +234,35 @@ func (runner *replicationJobRunner) run() {
...
+			delete(runner.failedJobs, existingFailedJob)
+		} else {
+			newFailedTimes := failedTimes + 1
+			runner.failedJobs[existingFailedJob] = newFailedTimes
+			if newFailedTimes > maxConsecutiveFailures {
+				maxConsecutiveFailures = newFailedTimes
+			}
+		}
+	}
+	for currentFailedJob, _ := range currentFailedJobs {
```
KiranRG: the second '_' is redundant .. in that you don't need it:
``` 
for currentFailedJob := range currentFailedJobs {
```

services/storehost/replicationJobRunner.go
```diff
@@ -224,9 +234,35 @@ func (runner *replicationJobRunner) run() {
...
+		if failedTimes, ok := runner.failedJobs[currentFailedJob]; !ok {
+			runner.failedJobs[currentFailedJob] = 1
+		} else {
+			runner.logger.WithFields(bark.Fields{
+				common.TagExt:  common.FmtExt(currentFailedJob),
+				`failed times`: failedTimes,
+			}).Info(`replication job failed for at least twice`)
```
KiranRG: it looks like this case is the same as the one in the else block at line 242 .. basically handling the intersection between runner.failedJobs and currentFailedJobs. You could probably move this log up there. Btw, I am wondering if you really need this log anyway .. since you are already logging when the error occurs. But I'll leave it to you.

kirg (approved):
Minor comments .. LGTM otherwise.

PR #88, opened 2017-03-08
ConfigMgr: Make the rule evaluator honor hierarchy

venkat1109:
Currently, the cassandra backed dynamic config manager has some assumptions about how wildcards are processed. Every config item handled by the configMgr is of the form:

serviceName.serviceVersion.sku.host.key=value

And as you go towards the right, the rules get more specific and the longest matching rule wins. However, with the current code, wildcards are *not* really fully supported. The configMgr assumes it will always be dealing with all the params i.e. {name,version,sku,host}.  

This patch does the following

* Makes the configMgr support wildcards fully
* Modifies rpm to truncate hostnames after the first period

### DIFF
--- a/common/dconfig/configmgr.go
+++ b/common/dconfig/configmgr.go
@@ -511,13 +511,15 @@ func newKVTreeNode() *kvTreeNode {
 
 func addToKVTree(root *kvTreeNode, version, sku, host, cfgKey, cfgValue string) {
 
-	var maxDepth = 1 // max depth of tree to traverse
+	var maxDepth int // max depth of tree to traverse
 
-	if sku != wildcardToken || host != wildcardToken {
-		maxDepth++
-		if host != wildcardToken {
-			maxDepth++
-		}
+	switch {
+	case host != wildcardToken:
+		maxDepth = 3
+	case sku != wildcardToken:
+		maxDepth = 2
+	default:
+		maxDepth = 1
 	}
 
 	curr := root
--- a/common/dconfig/configmgr.go
+++ b/common/dconfig/configmgr.go
@@ -438,7 +438,6 @@ func (cfgMgr *CassandraConfigManager) mkKVTreeForSvc(service string, items []*m.
 	return root
 }
 
-
 // mkConfig constructs a config object of given type using the
 // values from the given set of key,value strings. If the given
 // map of (k,v) is missing a config key, then this method will
--- a/common/dconfig/configmgr.go
+++ b/common/dconfig/configmgr.go
@@ -351,8 +351,8 @@ func (cfgMgr *CassandraConfigManager) mkKVTreeForSvc(service string, items []*m.
 
 	allSkus := make(map[string]struct{})
 	allVersions := make(map[string]struct{})
-	wildcardVersionItems := make([]*m.ServiceConfigItem, 0, len(items)) // items with version as wildcard
-	wildcardVersionSkuItems := make([]*m.ServiceConfigItem, 0, len(items)) // items with version as wildcard and sku not a wildcard
+	wildcardVersionItems := make([]*m.ServiceConfigItem, 0, len(items))              // items with version as wildcard
+	wildcardVersionSkuItems := make([]*m.ServiceConfigItem, 0, len(items))           // items with version as wildcard and sku not a wildcard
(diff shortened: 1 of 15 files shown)


common/dconfig/configmgr.go
```diff
@@ -339,46 +369,75 @@ func (cfgMgr *CassandraConfigManager) mkKVTreeForSvc(service string, items []*m.
 			continue
 		}
 
-		versions[version] = struct{}{}
+		if sku != wildcardToken && host != wildcardToken {
+			key := strings.Join([]string{service,version, sku, host,cfgKey}, ".")
```
KiranRG: cosmetic: the spaces between the items in the slice are inconsistent -- not sure why your editor/lint didn't fix it (mine did!)

common/dconfig/configmgr.go
```diff
@@ -339,46 +369,75 @@ func (cfgMgr *CassandraConfigManager) mkKVTreeForSvc(service string, items []*m.
...
+		for ver := range allVersions {
+			addToKVTree(root, ver, item.GetSku(), item.GetHostname(), item.GetConfigKey(), item.GetConfigValue())
+		}
+	}
 
-		for ver := range versions {
-			addToKVTree(root, ver, sku, host, cfgKey, cfgValue)
+	// apply svc.*. items as default for every svc.ver.*
+	for _, item := range wildcardVersionItems {
+ 		for ver := range allVersions {
```
KiranRG: cosmetic: there's a 'space' in the beginning of this line (that my editor) points out/fixes.

common/dconfig/configmgr.go
```diff
@@ -318,14 +325,37 @@ func (cfgMgr *CassandraConfigManager) mkConfigTreeForSvc(cfgType interface{}, de
...
 
 	root := newKVTreeNode()
 
-	versions := make(map[string]struct{})
-	wildcardVersionItems := make([]*m.ServiceConfigItem, 0, len(items))
+	allSkus := make(map[string]struct{})
+	allVersions := make(map[string]struct{})
+	wildcardVersionItems := make([]*m.ServiceConfigItem, 0, len(items)) // items with version as wildcard
+	wildcardVersionSkuItems := make([]*m.ServiceConfigItem, 0, len(items)) // items with version as wildcard and sku not a wildcard
+	versionToWildcardSkuItems := make(map[string][]*m.ServiceConfigItem, len(items)) // items with sku as wildcard
```
KiranRG: cosmetic: my editor/lint seems to align up the comments on these three lines to the same column.

common/dconfig/configmgr.go
```diff
@@ -455,7 +514,7 @@ func addToKVTree(root *kvTreeNode, version, sku, host, cfgKey, cfgValue string)
 
 	var maxDepth = 1 // max depth of tree to traverse
 
-	if sku != wildcardToken {
+	if sku != wildcardToken || host != wildcardToken {
 		maxDepth++
 		if host != wildcardToken {
 			maxDepth++
```
KiranRG:
```
switch {
case host != wildcardToken:
    maxDepth = 3
case sku != wildcardToken:
    maxDepth = 2
default:
    maxDepth = 1
}
```
^^ Clearer?

common/dconfig/configmgr.go
```diff
@@ -339,46 +369,75 @@ func (cfgMgr *CassandraConfigManager) mkKVTreeForSvc(service string, items []*m.
 			continue
 		}
 
-		versions[version] = struct{}{}
+		if sku != wildcardToken && host != wildcardToken {
+			key := strings.Join([]string{service,version, sku, host,cfgKey}, ".")
+			sku = wildcardToken
+			items[i].Sku = common.StringPtr(wildcardToken)
+			cfgMgr.logger.WithField(key, cfgValue).Warn("Forcing sku=*;host specific items MUST be sku agnostic")
```
KiranRG: It would be good to keep the log-field "tag" to be a constant string literal and move the variable/dynamic part to the "value" of the tag.

venkat1109: good idea, fixed this.

common/dconfig/configmgr.go
venkat1109: yeah.. editor wasn't running go fmt. Fixed this.

venkat1109:
This diff is too old now, closing in favor of #109

PR #89, opened 2017-03-08
Make CreateConsumerGroup work with StartFrom with any unit

KiranRG:
Since 'startFrom' does not specify a unit, make sure it works with any of the various units.

### DIFF
--- a/services/frontendhost/frontend.go
+++ b/services/frontendhost/frontend.go
@@ -301,31 +301,38 @@ func convertCGZoneConfigToInternal(cgZoneCfg *c.ConsumerGroupZoneConfig) *shared
 	return internalCGZoneCfg
 }
 
+// interpretTimeNanos converts the given timestamp to nanosecond units after
+// interpreting its units based on what yields the nearest time to 'now'.
+func interpretTimeNanos(ts int64) int64 {
+
+	// tsNanos, tsMicros, tsMillis and tsSeconds are what 'ts' would be
+	// would be if it were assumed to be in {nano,micro,milli,}seconds
+	// respectively and they are each converted to nanosecond units.  we
+	// use the timestamp that is closest to 'now' as the most reasonable
+	// interpretation of the given 'ts'. we can safely ignore potential
+	// overflows in our computations, because that would result in a
+	// negative number that would push it further from the 'now' timestamp.
+
+	tsNanos, tsMicros, tsMillis, tsSeconds := ts, ts*1e3, ts*1e6, ts*1e9
+
+	// 'now' in unix-nanos
+	now := time.Now().UnixNano()
+
+	// the time that is nearest to 'now' (in nanoseconds), would be the most
+	// reasonable interpretation of 'ts'
+	return common.FindNearestInt(now, tsNanos, tsMicros, tsMillis, tsSeconds)
+}
+
 // convertCreateCGRequestToInternal converts Cherami CreateConsumerGroupRequest to internal shared CreateConsumerGroupRequest
 func convertCreateCGRequestToInternal(createRequest *c.CreateConsumerGroupRequest) *shared.CreateConsumerGroupRequest {
 
-	// detect and correct the units for 'startFrom' (expected internally to be in nanoseconds)
+	// detect and correct the units for 'startFrom' (expected internally to be in nanoseconds);
+	// for '0'
 	startFrom := createRequest.GetStartFrom()
 
-	// Note: 2000-01-01T00:00 is 946684800 seconds since epoch
-	switch {
-	case startFrom == 0:
-		// special-case '0' and let it through
-
-	case startFrom < 946684800000:
-		// likely in seconds; convert to nanoseconds
-		startFrom *= 1000000000
-
-	case startFrom < 946684800000000:
-		// likely in milliseconds; convert to nanoseconds
-		startFrom *= 1000000
-
-	case startFrom < 946684800000000000:
-		// likely in microseconds; convert to nanoseconds
-		startFrom *= 1000
-
-	default:
-		// likely in nanoseconds already
+	// special-case a StartFrom of '0' (start from beginning)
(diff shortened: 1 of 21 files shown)


services/frontendhost/frontend.go
```diff
@@ -303,13 +303,38 @@ func convertCGZoneConfigToInternal(cgZoneCfg *c.ConsumerGroupZoneConfig) *shared
 
 // convertCreateCGRequestToInternal converts Cherami CreateConsumerGroupRequest to internal shared CreateConsumerGroupRequest
 func convertCreateCGRequestToInternal(createRequest *c.CreateConsumerGroupRequest) *shared.CreateConsumerGroupRequest {
+
+	// detect and correct the units for 'startFrom' (expected internally to be in nanoseconds)
+	startFrom := createRequest.GetStartFrom()
```
venkat1109: can you move this logic into a helper (that takes a int64 and returns nanos)... That way, you can unit test that function separately (preferably table-test)

services/frontendhost/frontend_test.go
```diff
@@ -663,6 +663,72 @@ func (s *FrontendHostSuite) TestFrontendHostCreateConsumerGroup() {
...
+// TestFrontendHostCreateConsumerGroupStartFrom tests whether CreateConsumerGroup is able to
+// detect and appropriately convert various startFrom units.
+func (s *FrontendHostSuite) TestFrontendHostCreateConsumerGroupStartFrom() {
+
+	testPath := uuid.New()
+	testCG := s.generateKey("/bar/CGName")
+	frontendHost, ctx := s.utilGetContextAndFrontend()
+
+	// set start-from to, say, four weeks ago
+	startFromTime := time.Now().Add(-4 * 7 * 24 * time.Hour)
```
venkat1109: need more test cases around boundary..

services/frontendhost/frontend_test.go
```diff
@@ -663,6 +663,72 @@ func (s *FrontendHostSuite) TestFrontendHostCreateConsumerGroup() {
...
+		frontendHost.writeCacheDestinationPathForUUID(destinationUUID(cgDesc.GetDestinationUUID()), testPath)
+
+		dlqPath, _ := common.GetDLQPathNameFromCGName(testCG)
+		destDesc := &shared.DestinationDescription{
+			Path:            common.StringPtr(dlqPath),
+			DestinationUUID: common.StringPtr(uuid.New()),
+		}
+
+		// create destination is needed because of dlq destination been created at time of consumer group creation
+		s.mockController.On("CreateDestination", mock.Anything, mock.Anything).Return(destDesc, nil).Run(func(args mock.Arguments) {
```
venkat1109: see comment above, don't need any of these for testing your helper

services/frontendhost/frontend.go
```diff
@@ -303,13 +303,38 @@ func convertCGZoneConfigToInternal(cgZoneCfg *c.ConsumerGroupZoneConfig) *shared
...
+
+	// Note: 2000-01-01T00:00 is 946684800 seconds since epoch
+	switch {
+	case startFrom == 0:
+		// special-case '0' and let it through
+
+	case startFrom < 946684800000:
```
thuningxu: Why not use common.MaxUint32 for seconds boundary? This goes to year 2038.

GuillaumeBailey: > Why not use common.MaxUint32 for seconds boundary? This goes to year 2038.

We have some values that are small, but valid. First is 0, which is a special value to mean 'now'. The second is that our documentation suggests that a very small time (e.g. '1') should be used to start from the oldest available messages. Since our CLI interprets its parameters as seconds, '1' second after epoch becomes '1E9' nanoseconds after epoch. We are currently at ~ 1.49E9 seconds after epoch, so for Kiran's scheme to work in the face of our CLI's deficiencies, we need to choose our cutoff points wisely.

Alternatively, we could also amend our CLI to treat small values (e.g. < 1E9) literally, rather than as seconds after epoch, which would simplify the model here.

services/frontendhost/frontend.go
KiranRG: Will do. Though I still think it is better to test the behaviour of "CreateConsumerGroup" itself .. not just the helper. In the future if, say, someone does away with calling the helper, the test (that only tests the helper) would still pass.

services/frontendhost/frontend_test.go
KiranRG: i am not sure what you mean by the boundary .. but i think we need to test whether this works okay with dates within "reasonable" boundaries. i'll add some more.

services/frontendhost/frontend_test.go
KiranRG: see my response above; i still think it is better to test at the CreateConsumerGroup API level.

services/frontendhost/frontend.go
KiranRG: I verified the behaviour with CLI. If *no* start-time is specified, it uses a StartFrom of "now". But if the user explicitly specifies "0", the CLI passed down "0".

I think we should fix our documentation to not say "1", but say "0" instead.

KiranRG:
Based on my discussion with @GuillaumeBailey, I liked his idea of using the interpretation of start-from that yields the time that is closest to "now", and use that. This change implements that.

services/frontendhost/frontend_test.go
```diff
@@ -663,6 +663,72 @@ func (s *FrontendHostSuite) TestFrontendHostCreateConsumerGroup() {
...
+
+		frontendHost.CreateConsumerGroup(ctx, req)
+	}
+
+	// test startFrom with varying units
+	testStartFrom(startFromNanos)
+	testStartFrom(startFromMicros)
+	testStartFrom(startFromMillis)
+	testStartFrom(startFromSeconds)
+	testStartFrom(0)
```
KiranRG: I need to understand this some more.

PR #91, opened 2017-03-09
Report per destination bytes in counters

aravindvs:
This patch makes sure we report the bytes in counters for the following:
* per destination
* per extent
* per host

We only report the delta from the last round and so we should reset it
once we report it to the controller.


venkat1109:
@aravindvs - Are these counts reported to the metrics system already ? If not, we should report it there as well.

aravindvs:
@venkat1109 some are getting reported but some aren't.. I will send out a separate patch to report to M3 on both input and output..

PR #92, opened 2017-03-11
report first/last addr/timestamp in replica-stats

KiranRG:
The diff adds support in storehost to report first/last timestamps and address (along with seqnum) with replica-stats.

This is more of a preview of the changes and is work-in-progress, since some queue-depth tests are failing against this change, that I am yet to root-cause/fix.


KiranRG:
I have cleaned up the code and split it into two separate pull-requests; and have closed out this one.

PR #94, opened 2017-03-13
Flatten tags for statsd in pre-defined order

### LINKED ISSUE #93: StatsdReporter merges tags in random order
In Uber metrics are tagged, but when they are converted to StatsD metrics, the tags are converted into prefixes. Unfortunately, because tags are `map[string]string`, during the conversion, we iterate this map in random order (in `statsdreporter.go`, `metricstoPrefix()`). This results in different prefixes being produced for the same metric, e.g.

image

thuningxu:
Fixes #93

### DIFF
--- a/common/metrics/statsdreporter.go
+++ b/common/metrics/statsdreporter.go
@@ -129,12 +129,14 @@ func mergeDicts(dic1 map[string]string, dic2 map[string]string) (resultDict map[
 
 // MetricWithPrefix is the default mapping for metrics to statsd keys.
 func metricstoPrefix(name string, tags map[string]string) string {
-	parts := []string{"cherami", name}
+	parts := []string{name}
 
-	for k := range tags {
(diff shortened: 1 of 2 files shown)


common/metrics/statsdreporter.go
```diff
@@ -127,14 +127,26 @@ func mergeDicts(dic1 map[string]string, dic2 map[string]string) (resultDict map[
 	return
 }
 
+// Cherami's metrics are tagged, but here we need to flatten 
+// it for statsd. We order tags by: Operation, Destination, 
+// ConsumerGroup, Hostname
+var tagsFlattenOrder = []string{
```
venkat1109: is this the right place for defining the possible tags and the order in which they appear ? 
What happens when someone introduces a new tag in metrics/defs.go ?

thuningxu: good point.

PR #96, opened 2017-03-14
Fix queue depth and related replica stats

KiranRG:
This diff contains two related changes:
1. Add/fix code in storehost to publish first/last message address and timestamps .. mostly intended for use for the queue-depth code
2. Fix queue-depth code and tests, that were breaking because of the fix above.

### DIFF
--- a/services/controllerhost/queueDepth.go
+++ b/services/controllerhost/queueDepth.go
@@ -41,10 +41,6 @@ import (
 )
 
 const (
-	// T471438, ADDR_SEAL is being written into replica_stats. We ignore small negative 26-bit 2's-compliment numbers,
-	// which are larger than this constant in signed 64-bit space
-	storageMetaValuesLimit = int64(1<<26) - (1 << 4)
-
 	// QueueDepthTabulationString can be added to a destination or CG owner email to request queue depth tabulation
 	// Note that Google allows something like this: gbailey+queueDepthTabulation@uber.com
 	// The above is still a valid email and will be delivered to gbailey@uber.com
--- a/services/controllerhost/queueDepth.go
+++ b/services/controllerhost/queueDepth.go
@@ -325,16 +325,15 @@ func (qdc *queueDepthCalculator) addExtentBacklog(
 
 func (qdc *queueDepthCalculator) computeBacklog(cgDesc *shared.ConsumerGroupDescription, cgExtent *metadata.ConsumerGroupExtent, storeMetadata *storeExtentMetadata, storeID string, logger bark.Logger) int64 {
 
-	var backlog int64 = 0 // default
+	var backlog int64 // backlog defaults to '0'
 	var iter = &qdc.iter
 
-	switch qdc.iter.isDLQ {
-	case true:
+	if qdc.iter.isDLQ {
 		// update backog, only if the begin/first seqnums are available
 		if storeMetadata.lastSequence != math.MaxInt64 && storeMetadata.beginSequence != math.MaxInt64 {
 			backlog = storeMetadata.lastSequence - storeMetadata.beginSequence + 1
 		}
-	case false:
+	} else {
 		// update backlog, only if there is an available seqnum
 		if storeMetadata.availableSequence != math.MaxInt64 {
 			backlog = storeMetadata.availableSequence - cgExtent.GetAckLevelSeqNo()
--- a/services/controllerhost/queueDepth.go
+++ b/services/controllerhost/queueDepth.go
@@ -41,6 +41,10 @@ import (
 )
 
 const (
+	// T471438, ADDR_SEAL is being written into replica_stats. We ignore small negative 26-bit 2's-compliment numbers,
+	// which are larger than this constant in signed 64-bit space
+	storageMetaValuesLimit = int64(1<<26) - (1 << 4)
+
 	// QueueDepthTabulationString can be added to a destination or CG owner email to request queue depth tabulation
 	// Note that Google allows something like this: gbailey+queueDepthTabulation@uber.com
 	// The above is still a valid email and will be delivered to gbailey@uber.com
@@ -321,15 +325,17 @@ func (qdc *queueDepthCalculator) addExtentBacklog(
 
 func (qdc *queueDepthCalculator) computeBacklog(cgDesc *shared.ConsumerGroupDescription, cgExtent *metadata.ConsumerGroupExtent, storeMetadata *storeExtentMetadata, storeID string, logger bark.Logger) int64 {
 
-	var backlog int64
+	var backlog int64 = 0 // default
 	var iter = &qdc.iter
 
 	switch qdc.iter.isDLQ {
 	case true:
+		// update backog, only if the begin/first seqnums are available
 		if storeMetadata.lastSequence != math.MaxInt64 && storeMetadata.beginSequence != math.MaxInt64 {
 			backlog = storeMetadata.lastSequence - storeMetadata.beginSequence + 1
 		}
 	case false:
+		// update backlog, only if there is an available seqnum
 		if storeMetadata.availableSequence != math.MaxInt64 {
 			backlog = storeMetadata.availableSequence - cgExtent.GetAckLevelSeqNo()
 		}
@@ -543,11 +549,12 @@ func (qdc *queueDepthCalculator) handleStartFrom(
 
 		// if doGaft == false, this just adjusts for retention, if applicable
 		if qualify {
-			trace += 100
 			consumerGroupExtent.WriteTime = common.Int64Ptr(int64(now))
 
 			// retention may have purged some messages, account for that
 			if storeMetadata.beginSequence != math.MaxInt64 {
+
+				trace += 100
 
 				// the 'ack-level' corresponds to message that has already been read (ie 'acked');
 				// so reduce one from the 'beginSequence' to since that msg has not been read.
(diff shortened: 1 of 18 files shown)


services/controllerhost/queueDepth.go
```diff
@@ -329,9 +326,13 @@ func (qdc *queueDepthCalculator) computeBacklog(cgDesc *shared.ConsumerGroupDesc
...
 	switch qdc.iter.isDLQ {
 	case true:
-		backlog = common.MaxInt64(0, storeMetadata.lastSequence-(storeMetadata.beginSequence+1))
+		if storeMetadata.lastSequence != math.MaxInt64 && storeMetadata.beginSequence != math.MaxInt64 {
+			backlog = storeMetadata.lastSequence - storeMetadata.beginSequence + 1
+		}
 	case false:
-		backlog = common.MaxInt64(0, storeMetadata.availableSequence-cgExtent.GetAckLevelSeqNo())
+		if storeMetadata.availableSequence != math.MaxInt64 {
+			backlog = storeMetadata.availableSequence - cgExtent.GetAckLevelSeqNo()
```
venkat1109: you want to still leave the common.MaxInt64(0, backlog) as is - these values are eventually consistent.. So, there is no strong guarantee that AckLevelSeqNo will be less than the availableSequence. And to be consistent, you can do the same thing for if (true) case

services/controllerhost/queueDepthCache.go
```diff
@@ -96,9 +95,6 @@ func (cache *storeExtentMetadataCache) fetchStoreExtentMetadata(extID string, st
 	}
 
 	rs := stats.GetReplicaStats()[0]
-	// store apparently reports illegal values
-	// fix them if needed
-	cache.fixReplicaStatsIfBroken(rs)
```
venkat1109: still want to leave this around to be backward compatible.. We should be able to delete this code once all the data has been touched by your new reporting.

services/storehost/extMgr.go
```diff
@@ -783,39 +776,57 @@ func (ext *extentContext) delete() {
...
+func (ext *extentContext) initFirstMsg() error {
 
 	// FIXME: for timer-queues, this would return the seqNum of the first
 	// _available_ message to deliver, and *not* the first enqueued message.
 
 	_, key, err := ext.store.SeekFirst()
 
-	if err == nil && key != storage.InvalidKey {
-		_, beginSeqNum = ext.deconstructKey(key)
+	if err == nil && key != storage.InvalidKey &&
```
venkat1109: positive assertions are more readable then negative assertions. In addition,
https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

So, prefer

if err !=nil || key == storage.InvalidKey || ext.isSealExtentKey(key) {
     // init
    return
}

services/storehost/extMgr.go
```diff
@@ -783,39 +776,57 @@ func (ext *extentContext) delete() {
...
 	// FIXME: for timer-queues, this would return the seqNum of the last
 	// _available_ message to deliver, and *not* the last enqueued message.
 
-	_, key, err := ext.store.SeekLast()
+	// find the last message (excluding any seal-key markers)
+	lastPossibleKey := ext.constructSealExtentKey(0) - 1
+
+	_, key, err := ext.store.SeekFloor(lastPossibleKey)
 
 	if err == nil && key != storage.InvalidKey {
```
venkat1109: see comment above about control-flow

services/storehost/inConn.go
```diff
@@ -303,7 +304,7 @@ func (t *inConn) writeMessagesPumpAppendOnly(msgC <-chan *inMessage, ackC chan<-
 			}
 
 			// if seqNum is not in expected (continuously increasing) order, log error as FYI
-			if x.getLastSeqNum()+1 != msgSeqNum && x.getLastSeqNum() != -1 {
+			if x.getLastSeqNum()+1 != msgSeqNum && x.getLastSeqNum() != math.MaxInt64 {
```
venkat1109: You might be better off defining a const for invalid first/last {seq,addr,ts} values. const helps if (1) ever you change this value to something else other than math.MaxInt64 (2) Finding out all references where invalid_ values are accessed in code.

services/controllerhost/queueDepth.go
```diff
@@ -544,10 +545,17 @@ func (qdc *queueDepthCalculator) handleStartFrom(
 		if qualify {
 			trace += 100
 			consumerGroupExtent.WriteTime = common.Int64Ptr(int64(now))
-			consumerGroupExtent.AckLevelSeqNo = common.Int64Ptr(common.MaxInt64(
-				storeMetadata.beginSequence, // Retention may have removed some messages
-				int64(startFromSeq),         // Otherwise, act like we had just opened this extent at startFrom, i.e. don't count messages before startFrom
-			))
+
+			// retention may have purged some messages, account for that
+			if storeMetadata.beginSequence != math.MaxInt64 {
```
venkat1109: (1) this changes behavior, you still want to update AckLevelSeqNo if quality=true
(2) trace logging is broken
(3) Needs to be backwards compatible (i.e. also deal with beginSeq=-1)

services/controllerhost/queueDepth.go
```diff
@@ -329,9 +326,13 @@ func (qdc *queueDepthCalculator) computeBacklog(cgDesc *shared.ConsumerGroupDesc
 
 	switch qdc.iter.isDLQ {
 	case true:
-		backlog = common.MaxInt64(0, storeMetadata.lastSequence-(storeMetadata.beginSequence+1))
+		if storeMetadata.lastSequence != math.MaxInt64 && storeMetadata.beginSequence != math.MaxInt64 {
+			backlog = storeMetadata.lastSequence - storeMetadata.beginSequence + 1
```
venkat1109: breaks backward compatibility

services/storehost/storehost.go
```diff
@@ -789,21 +785,21 @@ func (t *StoreHost) SealExtent(ctx thrift.Context, req *store.SealExtentRequest)
 		// of 'seqNumUnspecifiedSeal' to go through, to support sealing an extent as a best effort.
 
 		switch {
-		case lastSeqNum == -1 && unspecifiedSealSeqNum: // no sealSeqNum specified and we don't know lastSeqNum
+		case lastSeqNum == math.MaxInt64 && unspecifiedSealSeqNum: // no sealSeqNum specified and we don't know lastSeqNum
```
venkat1109: this change isn't backward compatible

test/integration/integration_test.go
```diff
@@ -2386,15 +2387,19 @@ func (s *NetIntegrationSuiteParallelB) TestQueueDepth() {
 				stats := srs.GetExtent().GetReplicaStats()[0] // Modify the existing stats. We would otherwise write nils to most fields
 
 				if reset { // clearing
-					if bs > 0 {
+					if bs != math.MaxInt64 {
```
venkat1109: compatibility

services/storehost/inConn.go
KiranRG: Done.

services/controllerhost/queueDepth.go
```diff
@@ -324,14 +325,20 @@ func (qdc *queueDepthCalculator) addExtentBacklog(
 
 func (qdc *queueDepthCalculator) computeBacklog(cgDesc *shared.ConsumerGroupDescription, cgExtent *metadata.ConsumerGroupExtent, storeMetadata *storeExtentMetadata, storeID string, logger bark.Logger) int64 {
 
-	var backlog int64
+	var backlog int64 = 0 // default
```
venkat1109: initialization is redundant. Defaults to zero anyways.

services/controllerhost/queueDepth.go
```diff
@@ -542,12 +549,27 @@ func (qdc *queueDepthCalculator) handleStartFrom(
...
+
+			// retention may have purged some messages, account for that
+			if storeMetadata.beginSequence != math.MaxInt64 {
+
+				trace += 100
+
+				// the 'ack-level' corresponds to message that has already been read (ie 'acked');
+				// so reduce one from the 'beginSequence' to since that msg has not been read.
+				consumerGroupExtent.AckLevelSeqNo = common.Int64Ptr(common.MaxInt64(
+					storeMetadata.beginSequence-1, // Retention may have removed some messages
```
venkat1109: can the beginSequence be 0 ?

services/controllerhost/queueDepth.go
```diff
@@ -324,14 +325,20 @@ func (qdc *queueDepthCalculator) addExtentBacklog(
...
-		backlog = common.MaxInt64(0, storeMetadata.lastSequence-(storeMetadata.beginSequence+1))
+		// update backog, only if the begin/first seqnums are available
+		if storeMetadata.lastSequence != math.MaxInt64 && storeMetadata.beginSequence != math.MaxInt64 {
+			backlog = storeMetadata.lastSequence - storeMetadata.beginSequence + 1
+		}
 	case false:
-		backlog = common.MaxInt64(0, storeMetadata.availableSequence-cgExtent.GetAckLevelSeqNo())
+		// update backlog, only if there is an available seqnum
+		if storeMetadata.availableSequence != math.MaxInt64 {
+			backlog = storeMetadata.availableSequence - cgExtent.GetAckLevelSeqNo()
```
venkat1109: I believe you still need max(0, diff) for backwards compatibility. If I understand right, the availableSequence can be -1 for empty extents with the old code.

services/controllerhost/queueDepth.go
KiranRG: No. Message seq-nums start from "1". If there are no messages in the extent, then the beginSequence would be "MaxInt64".

services/controllerhost/queueDepth.go
KiranRG: The 'fixReplicaStatsIfBroken' is expected to "fix" such values (those generated from the old behaviour) .. and make them "MaxInt64" (as is expected from the new behaviour).

services/controllerhost/queueDepth.go
KiranRG: I put that to make it explicit that we rely on a default of "0". I'll probably just leave a comment there that says this.

PR #97, opened 2017-03-15
multi zone consumer group: cg metadata propagation(fast and slow path)

datoug:
This diff adds the multi_zone cg metadata propagation. Here's a summary of the change
1. Move DLQ creation from frontend to metadata. This is because replicator also needs to create cg by calling metadata so all logic needs to happen in metadata
2. cg CUD fast path. Request flow: local frontend->local controller (will call metadata to create cg locally)->local replicator->remote replicator(will call metadata to create cg locally)
3. cg CUD slow path: replicator in each zone will talk to authoritative zone, and compare (update is not in this diff because the current diff is too large now)

related thrift change: https://github.com/uber/cherami-thrift/pull/12

@GuillaumeBailey you're only required to review change #1 (change in metadata_cassandra) but feel free to review the whole PR

### DIFF
--- a/clients/metadata/metadata_cassandra.go
+++ b/clients/metadata/metadata_cassandra.go
@@ -590,7 +590,7 @@ func unmarshalDstZoneConfigs(configsData []map[string]interface{}) []*shared.Des
 // ReadDestination implements the corresponding TChanMetadataServiceClient API
 // Either path or destinationUUID can be specified.
 // Deleted destinations are returned with DELETED status only when destinationUUID is used.
-func (s *CassandraMetadataService) ReadDestination(ctx thrift.Context, getRequest *m.ReadDestinationRequest) (result *shared.DestinationDescription, err error) {
+func (s *CassandraMetadataService) ReadDestination(ctx thrift.Context, getRequest *shared.ReadDestinationRequest) (result *shared.DestinationDescription, err error) {
 	var query *gocql.Query
 	var uuid, sql string
 	if getRequest.Path != nil {
@@ -678,7 +678,7 @@ func (s *CassandraMetadataService) UpdateDestination(ctx thrift.Context, updateR
 			Message: "UpdateDestination: Update to DELETED status is not allowed. Use DeleteDestination instead.",
 		}
 	}
-	getDestination := &m.ReadDestinationRequest{
+	getDestination := &shared.ReadDestinationRequest{
 		DestinationUUID: common.StringPtr(updateRequest.GetDestinationUUID()),
 	}
 	existing, err := s.ReadDestination(nil, getDestination)
@@ -780,7 +780,7 @@ func (s *CassandraMetadataService) UpdateDestination(ctx thrift.Context, updateR
 
 // DeleteDestination implements the corresponding TChanMetadataServiceClient API
 func (s *CassandraMetadataService) DeleteDestination(ctx thrift.Context, deleteRequest *shared.DeleteDestinationRequest) error {
-	getDestination := &m.ReadDestinationRequest{
+	getDestination := &shared.ReadDestinationRequest{
 		Path: common.StringPtr(deleteRequest.GetPath()),
 	}
 	existing, err := s.ReadDestination(nil, getDestination)
@@ -834,7 +834,7 @@ func (s *CassandraMetadataService) DeleteDestinationUUID(ctx thrift.Context, del
 		return &shared.BadRequestError{Message: "Missing UUID param"}
 	}
 
-	getDestination := &m.ReadDestinationRequest{
+	getDestination := &shared.ReadDestinationRequest{
 		DestinationUUID: common.StringPtr(deleteRequest.GetUUID()),
 	}
 	existing, err := s.ReadDestination(nil, getDestination)
@@ -1003,7 +1003,7 @@ func (s *CassandraMetadataService) ListDestinationsByUUID(ctx thrift.Context, li
 		*d.DLQMergeBefore = int64(cqlTimestampToUnixNano(*d.DLQMergeBefore))
 
 		if listRequest.GetValidateAgainstPathTable() {
-			readRequest := &m.ReadDestinationRequest{
+			readRequest := &shared.ReadDestinationRequest{
 				Path: d.Path,
 			}
 
@@ -1223,7 +1223,7 @@ func (s *CassandraMetadataService) CreateConsumerGroupUUID(ctx thrift.Context, r
 			switch err.(type) {
 			case *shared.EntityAlreadyExistsError:
 				log.WithFields(log.Fields{common.TagCnsm: common.FmtCnsm(cgUUID)}).Info("DeadLetterQueue destination already existed")
-				mDLQReadRequest := m.ReadDestinationRequest{
+				mDLQReadRequest := shared.ReadDestinationRequest{
 					Path: dlqCreateRequest.Path,
 				}
 
@@ -1245,7 +1245,7 @@ func (s *CassandraMetadataService) CreateConsumerGroupUUID(ctx thrift.Context, r
 		log.WithFields(log.Fields{common.TagCnsm: common.FmtCnsm(cgUUID)}).Info("DeadLetterQueue destination not being created")
 	}
 
-	dstInfo, err := s.ReadDestination(nil, &m.ReadDestinationRequest{Path: common.StringPtr(createRequest.GetDestinationPath())})
+	dstInfo, err := s.ReadDestination(nil, &shared.ReadDestinationRequest{Path: common.StringPtr(createRequest.GetDestinationPath())})
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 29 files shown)


services/replicator/metadataReconciler.go
```diff
@@ -119,13 +119,19 @@ func (r *metadataReconciler) run() {
...
 	}
 
-	// destination metadata reconciliation is only needed if this is a non-authoritative zone
+	// destination/cg metadata reconciliation is only needed if this is a non-authoritative zone
 	if r.localZone != r.replicator.getAuthoritativeZone() {
 		r.m3Client.UpdateGauge(metrics.ReplicatorReconcileScope, metrics.ReplicatorReconcileDestRun, 1)
-		err = r.reconcileDestMetadata()
+		localDests, remoteDests, err := r.reconcileDestMetadata()
 		if err != nil {
 			r.m3Client.UpdateGauge(metrics.ReplicatorReconcileScope, metrics.ReplicatorReconcileDestFail, 1)
```
venkat1109: missing returns statement ?

services/frontendhost/frontend.go
```diff
@@ -1033,51 +1028,6 @@ func (h *Frontend) CreateConsumerGroup(ctx thrift.Context, createRequest *c.Crea
 		return nil, err
 	}
 
-	// Dead Letter Queue destination creation
-
-	// Only non-UUID (non-DLQ) destinations get a DLQ for the corresponding consumer groups
-	// We may create a consumer group consume a DLQ destination and no DLQ destination creation needed in this case
-	if common.PathRegex.MatchString(createRequest.GetDestinationPath()) {
```
venkat1109: where is this moved ? Can't figure out from this diff.

services/replicator/metadataReconciler.go
```diff
@@ -265,6 +286,121 @@ func (r *metadataReconciler) reconcileDest(localDests []*shared.DestinationDescr
 	r.m3Client.UpdateGauge(metrics.ReplicatorReconcileScope, metrics.ReplicatorReconcileDestFoundMissing, replicatorReconcileDestFoundMissingCount)
 }
 
+func (r *metadataReconciler) reconcileCg(localCgs []*shared.ConsumerGroupDescription, remoteCgs []*shared.ConsumerGroupDescription) {
+	var replicatorReconcileCgFoundMissingCount int64
+	localCgsSet := make(map[string]*shared.ConsumerGroupDescription)
```
venkat1109: make it the same length as len(localCgs) to avoid reallocations.

services/replicator/metadataReconciler.go
```diff
@@ -265,6 +286,121 @@ func (r *metadataReconciler) reconcileDest(localDests []*shared.DestinationDescr
...
+				} else {
+					lclLg.Info(`Found ConsumerGroup is deleted in both remote and local`)
+					continue
+				}
+				continue
+			}
+
+			// TODO case #2: ConsumerGroup exists in both remote and local, try to compare the property to see if anything gets updated
+		} else {
+			// case #3: ConsumerGroup exists in remote, but not in local. Create the ConsumerGroup locally
```
venkat1109: case #4: missing remote side, but exist locally ?

venkat1109: [ Not a Blocker ] Given that this is a state machine, one way to avoid nested if else is to use a table driven approach of state handlers. For example, if you have a table like the following:

stateHandlers[localCG.Status][remoteCG.status] -> func(localCG, remoteCG)

Then, you just have to index into this table to invoke the appropriate handler for the states.

datoug: for case #4, since we have large TTL(30 days) for the uuid table so it's unlikely we actually need #4 (it should be covered in case #1 (ConsumerGroup gets deleted in remote, but not deleted in local. Delete the ConsumerGroup locally).

Same behavior applies to destination reconciliation.

The extent reconciliation, we do have case 4 because the ttl is 1 day.

I'll add some comment to clarify this.

clients/metadata/metadata_cassandra.go
```diff
@@ -1181,19 +1186,73 @@ const (
...
+	// Dead Letter Queue destination creation
+
+	// Only non-UUID (non-DLQ) destinations get a DLQ for the corresponding consumer groups
+	// We may create a consumer group consume a DLQ destination and no DLQ destination creation needed in this case
+	var dlqUUID *string
+	if common.PathRegex.MatchString(createRequest.GetDestinationPath()) {
+		dlqCreateRequest := shared.NewCreateDestinationRequest()
+		dlqCreateRequest.ConsumedMessagesRetention = common.Int32Ptr(defaultDLQConsumedRetention)
+		dlqCreateRequest.UnconsumedMessagesRetention = common.Int32Ptr(defaultDLQUnconsumedRetention)
+		dlqCreateRequest.OwnerEmail = common.StringPtr(defaultDLQOwnerEmail)
```
venkat1109: use the same owner email in the CreateConsumerGroupRequest.

clients/metadata/metadata_cassandra.go
```diff
@@ -1181,19 +1186,73 @@ const (
...
-func (s *CassandraMetadataService) CreateConsumerGroup(ctx thrift.Context, request *shared.CreateConsumerGroupRequest) (*shared.ConsumerGroupDescription, error) {
+func (s *CassandraMetadataService) CreateConsumerGroupUUID(ctx thrift.Context, request *shared.CreateConsumerGroupUUIDRequest) (*shared.ConsumerGroupDescription, error) {
+	createRequest := request.GetRequest()
+	cgUUID := request.GetConsumerGroupUUID()
+
+	// Dead Letter Queue destination creation
+
+	// Only non-UUID (non-DLQ) destinations get a DLQ for the corresponding consumer groups
+	// We may create a consumer group consume a DLQ destination and no DLQ destination creation needed in this case
+	var dlqUUID *string
```
venkat1109: This method is too long now. Can you move the dlq creation logic into its method ?

clients/metadata/metadata_cassandra_test.go
```diff
@@ -1727,8 +1726,17 @@ func (s *CassandraSuite) TestConsumerGroupCRUD() {
...
+	expectedCG.DeadLetterQueueDestinationUUID = common.StringPtr(gotCG.GetDeadLetterQueueDestinationUUID())
 	assertConsumerGroupsEqual(s, expectedCG, gotCG)
 
+	// make sure the DLQConsumerGroupUUID from DLQ dest is correct
+	dlqDestReq := &shared.ReadDestinationRequest{
+		Path: common.StringPtr(gotCG.GetDeadLetterQueueDestinationUUID()),
+	}
+	dlqDest, err := s.client.ReadDestination(nil, dlqDestReq)
+	assert.Nil(err, "Read Dlq destination failed")
+	assert.Equal(dlqDest.GetDLQConsumerGroupUUID(), gotCG.GetConsumerGroupUUID())
```
venkat1109: can you also assert that the retention values are as expected ?

services/replicator/metadataReconciler.go
```diff
@@ -262,9 +284,130 @@ func (r *metadataReconciler) reconcileDest(localDests []*shared.DestinationDescr
...
+			if remoteCg.GetStatus() == shared.ConsumerGroupStatus_DELETED {
+				lclLg.Info(`Found missing ConsumerGroup from remote but in deleted state`)
+				continue
+			}
+
+			destDesc, err := r.readDestinationInAuthoritativeZone(remoteCg.GetDestinationUUID())
+			if err != nil {
+				lclLg.WithFields(bark.Fields{
+					common.TagErr: err,
+				}).Error(`Failed to create ConsumerGroup in local zone because read destination failed`)
```
venkat1109: readDestination failed in remote zone.

services/replicator/replicator.go
```diff
@@ -568,31 +568,280 @@ func (r *Replicator) deleteDestinationRemoteCall(zone string, logger bark.Logger
...
+		common.TagDstPth: common.FmtDstPth(updateRequest.GetDestinationPath()),
+		common.TagDst:    common.FmtDst(cgDesc.GetDestinationUUID()),
+		common.TagDLQID:  common.FmtDLQID(cgDesc.GetDeadLetterQueueDestinationUUID()),
+		`IsMultiZone`:    cgDesc.GetIsMultiZone(),
+		`ActiveZone`:     cgDesc.GetActiveZone(),
+	}).Info(`Updated cg`)
+	return cgDesc, nil
 }
 
 // UpdateRemoteConsumerGroup propagate update to multiple remote zones, expect to be called by local zone services
```
venkat1109: `expect to be called by local zone services`: Does the comment imply that a non-authoritative zone can update a consumer group ?

datoug: the plan is to limit destination/cg CRUD operation in authoritative zone only, although we haven't implemented that yet. I think 'local zone' here implies authoritative zone, I've added some comment to clarify

PR #101, opened 2017-03-24
Start kafka in travis

boy-uber:
Start zookeeper/kafka cluster in .travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,9 +51,6 @@ before_install:
   - mkdir -p kafka && tar xzf kafka.tgz -C kafka --strip-components 1
   - nohup bash -c "cd kafka && bin/zookeeper-server-start.sh config/zookeeper.properties &"
   - nohup bash -c "cd kafka && bin/kafka-server-start.sh config/server.properties &"
-  - sleep 10
-  - kafka/bin/kafka-topics.sh --create --partitions 2 --replication-factor 1 --topic test_topic_1 --zookeeper localhost:2181
-  - kafka/bin/kafka-topics.sh --zookeeper localhost:2181 --list
 
 install:
   - go get -u github.com/Masterminds/glide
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -47,6 +47,13 @@ before_install:
   - pip install --user ccm
   - pip install --user cqlsh==5.0.3
   - sudo apt-get install gcc g++ -y
+  - wget http://www.us.apache.org/dist/kafka/0.10.0.1/kafka_2.11-0.10.0.1.tgz -O kafka.tgz
+  - mkdir -p kafka && tar xzf kafka.tgz -C kafka --strip-components 1
+  - nohup bash -c "cd kafka && bin/zookeeper-server-start.sh config/zookeeper.properties &"
+  - nohup bash -c "cd kafka && bin/kafka-server-start.sh config/server.properties &"
+  - sleep 10
```
boy-uber: I see. Let check whether there is a better way :)

thuningxu: There is no reason to sleep here. The build process takes many minutes. By the time the build finishes kafka should be up already.

.travis.yml
```diff
@@ -47,6 +47,13 @@ before_install:
...
+  - kafka/bin/kafka-topics.sh --create --partitions 2 --replication-factor 1 --topic test_topic_1 --zookeeper localhost:2181
+  - kafka/bin/kafka-topics.sh --zookeeper localhost:2181 --list
```
thuningxu: Not sure why we need to do it here. Move to ##script## section (after ##make##)?

.travis.yml
boy-uber: Sleep here is to make sure the following topic creation could succeed (after kafka server starts and runs).

.travis.yml
boy-uber: I decided to totally remove the topic creation, since the kafka cluster will auto create topic.

boy-uber:
I updated the code and removed sleep there, also remove the topic creation/list script since topic will be auto created.

boy-uber:
Merge to master: start kafka cluster in travis CI

PR #105, opened 2017-03-24
multi zone cg extent metadata propagation

datoug:
This patch adds the cg extent metadata propagation for multi_zone cg(includes both fast and slow path).

On the fast path:
1. when a cg extent is created in controller, the request will be forwarded to local replicator, then to remote replicator, and finally to remote controller, which will then create the cg in remote zone
2. when ack level is updated in output host, the request will be forwarded to local replicator, then to remote replicator, which will then update the ack level in remote zone

On the slow path, we'll compare the list of cg extent with every other zone. And
1. If the extent is missing, create it locally
2. If the extent status is "behind", update the status locally
3. If the ack level is behind, update the ack level locally


services/replicator/replicator.go
```diff
@@ -932,6 +932,215 @@ func (r *Replicator) CreateRemoteExtent(ctx thrift.Context, createRequest *share
...
+	// acquire remote zone replicator thrift client
+	client, err := r.replicatorclientFactory.GetReplicatorClient(zone)
+	if err != nil {
+		r.m3Client.IncCounter(metrics.ReplicatorSetAckOffsetInRemoteScope, metrics.ReplicatorFailures)
+		logger.WithField(common.TagErr, err).Error(`Get remote replicator client failed`)
+		return err
+	}
+
+	// send to remote zone replicator
+	ctx, cancel := thrift.NewContext(remoteReplicatorCallTimeOut)
```
venkat1109: consider making this timeout value same as the output host ack offset flush interval (which is 5 seconds, I think)
there is no point in blocking on this call, if the data becomes stale every 5 seconds.

PR #107, opened 2017-03-27
outputhost: fix a bug that results in stuck consumer groups

venkat1109:
Uncovered a bug in output host while investigating a customer reported case last week. The symptom is a stuck consumer group i.e. consumers don't make any progress, no messages are sent from output hosts to clients. 

The trigger for this bug is one of the consuming worker process on the application rebooting, thereby killing all connections from that host to the output host. When that happens, there is a chance of a message getting black-holed per connection on the output host i.e. the last message sent on the socket from the output host to client can be forgotten by the output host. This message won't be acked by the worker process, because the socket is closed. It won't be redelivered by output-host because it lost the state for it. As a result, there will be a hold in in the sequence number space, which will never get filled and the ConsumerGroup will be stuck waiting for that hole to be filled.

The root cause is the output host consConnection failing to put a message back to msgDeliveryCache during connection shutdown. This patch does the following:

* Fixes the bug
* Adds a unit test covering the case that triggered the bug
* While working on the fix, noticed a unit test, which was totally broken (it was a no-op). Fixed this unit test - TestOutputHostReadMessage

### DIFF
--- a/services/outputhost/consconnection.go
+++ b/services/outputhost/consconnection.go
@@ -58,9 +58,10 @@ type (
 		// one message at a time) and are read during connection close after
 		// the pumps have closed -- therefore these don't need to be protected
 		// for concurrent access.
-		recvCreds  int64 // total credits received
-		sentMsgs   int64 // total messages sent out
-		reSentMsgs int64 // count of sent messages that were re-deliveries
+		recvCreds      int64 // total credits received
+		sentMsgs       int64 // total messages sent out
+		reSentMsgs 	   int64 // count of sent messages that were re-deliveries
+		sentToMsgCache int64 // count of message sent to message cache
 
 		lk     sync.Mutex
 		opened bool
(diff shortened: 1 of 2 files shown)


services/outputhost/consconnection.go
```diff
@@ -408,6 +399,15 @@ func (conn *consConnection) createMsgAndWriteToClientUtil(msg *cherami.ConsumerM
...
 		}
 	}
+	// Always add to the messageCache regardless of whether the write
+	// to the client succeeded or not. If we don't add it to the
+	// message cache, the message is *lost* and will never be redelivered.
+	// This can cause the CG to be stuck forever waiting for an ACK from client.
+	// The blocking operation below can delay the closing of stream when a
+	// conn close signal is received, the assumption here is that the
+	// operation if it blocks, will only do so for less than a second
+	msgCacheCh <- cacheMsg{msg: msg, connID: conn.connID}
```
aravindvs: This is a valid assumption but we should add an alert on message cache layer if any of the "cases" there takes more than a second. We can do it as a followup patch. @GuillaumeBailey already measures the time it takes on each "case" there.

PR #110, opened 2017-03-28
Add graceful drain of extents on inputhost

aravindvs:
This patch simply implements the graceful drain on inputhost without any client side involvement.
Whenever an inputhost gets a graceful drain request, it
does the following:
(1) Marks the extent as draining and stops the writepump and waits
    for the write pump to drain (for a timeout period)
(2) On the read pump if we have drained all the messages, then we
    can safely close the extent and go away.

A couple of caveats:
1. Note: close is idempotent and so we should be able to force close
   an extent anytime during the drain process.
2. We just need the RLock() during read pump aggregation and since it
   is not contended the profiling of inputhost didn't reveal it to be
   at top.

Also add a unit test for the same.

### DIFF
--- a/services/inputhost/exthost.go
+++ b/services/inputhost/exthost.go
@@ -110,6 +110,8 @@ type (
 
 		maxSizeBytes  int64 // max size for this extent; TODO: Make this dynamically configurable?
 		currSizeBytes int64 // current size of this extent
+
+		state extHostState // state of this extHost
 	}
 
 	// Holds a particular extent for use by multiple publisher connections.
@@ -136,8 +138,13 @@ type (
 		sendTimer *common.Timer
 	}
 
-	extCacheClosedCb   func(string)
-	extCacheUnloadedCb func(string)
+	extHostState int
+)
+
+const (
+	extHostActive extHostState = iota
+	extHostDraining
+	extHostInactive
 )
 
 const (
@@ -279,10 +286,8 @@ func (conn *extHost) close() {
 	// 1. stop the write pump to replicas and wait for the pump to close
 	// 2. close the replica streams
 	// 3. stop the read pump from replicas
-	close(conn.closeChannel)
-	if ok := common.AwaitWaitGroup(&conn.waitWriteWG, defaultWGTimeout); !ok {
-		conn.logger.Fatal("waitWriteGroup timed out")
-	}
+	conn.stopWritePump()
+
 	for _, stream := range conn.streams {
 		stream.conn.close()
 		// stop the timer as well so that it gets gc'ed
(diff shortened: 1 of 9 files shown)


services/inputhost/pathCache.go
```diff
@@ -508,3 +508,16 @@ func (pathCache *inPathCache) getState() *admin.DestinationState {
 
 	return destState
 }
+
+func (pathCache *inPathCache) drainExtent(extUUID string) error {
+	pathCache.RLock()
+	defer pathCache.RUnlock()
+
+	extCache, ok := pathCache.extentCache[extentUUID(extUUID)]
```
venkat1109: You might have already thought through this and may be some of the following things are planned for another diff. But questions for me:
(1) shouldn't there by a drain in pubConnection as well ?
(2) If you have (1), do you still need a drain in the extHost as well  ? 
(3) If you have drain in both pubConn & extHost, don't you have to call pubConn.drain() for all conns, wait before calling extHost.drain() ?

services/inputhost/exthost.go
```diff
@@ -853,3 +865,49 @@ func (conn *extHost) getState() *admin.InputDestExtent {
...
+}
+
+// isDrained returns true if the following conditions are true:
+// 	(1) extHost state is "extHostDraining"
+//	(2) we have seen the reply for the last sent sequence number
+// if not, it returns false
+// Note: RLock() is sufficient here.
+func (conn *extHost) isDrained(replySeqNo int64) bool {
+	conn.lk.RLock()
+	if conn.state != extHostDraining {
```
aravindvs: This is mainly to "unblock" the readPump once it has aggregated everything. So I would call it isDrained(). ideally if conn.state is inactive, i don't even want to do anything here because at that point we would have closed everything and there is no point in checking the seq no stuff. In summary this should be checked only if the state is in "Draining".. other stuff can be ignored since we are hard failing anyway.

services/inputhost/pathCache.go
aravindvs: Yeah i am sending the next PR which will drain the pubconnection appropriately. I need to do (3) and yes i have a guarantee to do that. 

And note, these are all best effort so if a connection is stuck for some reason (say a socket Write is blocking, I don't want to wait there forever).

services/inputhost/pathCache.go
```diff
@@ -508,3 +508,16 @@ func (pathCache *inPathCache) getState() *admin.DestinationState {
...
+	if !ok {
+		return &cherami.EntityNotExistsError{}
+	}
+
+	go extCache.connection.drain()
```
aravindvs: No.. I didn't have it planned that way, the timeouts are enforced on the drain layers and the controller will get an ExtentUnreachableNotification() call or a node failed event if it crashes.

aravindvs:
Abandoning this in favor of https://github.com/uber/cherami-server/pull/112

PR #112, opened 2017-03-29
Graceful drain on inputhost along with client drain

aravindvs:
This patch simply adds support to drain connections as well.
* To make sure we do this correctly, we maintain a "WritableExtents"
metric which is slightly different from the "OpenExtents" because
a drained extent is still open but is read only.
* When the number of WritableExtents goes to zero, we should send a
DRAIN command on all the connections.
* The connections will simply wait for a timeout period for the drain to
finish. If things finish successfully, we will unload the pathCache
and close all connections automatically. If not, the timeout period will
hit and we will drop the connection object.
* Add an integration test to test this flow.

### DIFF
--- a/services/inputhost/exthost.go
+++ b/services/inputhost/exthost.go
@@ -282,6 +282,10 @@ func (conn *extHost) close() {
 
 	conn.closed = true
 
+	// make sure the number of writable extents is updated just once
+	if conn.state == extHostActive {
+		conn.dstMetrics.Decrement(load.DstMetricNumWritableExtents)
+	}
 	// Shutdown order:
 	// 1. stop the write pump to replicas and wait for the pump to close
 	// 2. close the replica streams
@@ -900,14 +904,27 @@ func (conn *extHost) isDrained(replySeqNo int64) bool {
 	return ret
 }
 
+// prepDrain just decrements the number of writable extents for this destination
+func (conn *extHost) prepDrain() bool {
+	conn.lk.Lock()
+	if conn.state != extHostActive {
+		conn.lk.Unlock()
+		return false
+	}
+	conn.dstMetrics.Decrement(load.DstMetricNumWritableExtents)
+	conn.lk.Unlock()
+	return true
+}
+
 // drain simply stops the write pump and marks the state as such
 func (conn *extHost) drain() error {
 	conn.lk.Lock()
-	defer conn.lk.Unlock()
 	if conn.state != extHostActive {
+		conn.lk.Unlock()
 		// already drained and is not active
 		return nil
 	}
 	conn.stopWritePump()
+	conn.lk.Unlock()
 	return nil
 }
--- a/services/inputhost/exthost.go
+++ b/services/inputhost/exthost.go
@@ -142,9 +142,10 @@ type (
 )
 
 const (
-	extHostActive extHostState = iota
-	extHostDraining
-	extHostInactive
+	extHostActive       extHostState = iota
+	extHostDrainPrepped              // used by the pathcache to determine if the connections need to be torn down or not
+	extHostDraining                  // used when the actual drain is happening
+	extHostInactive                  // used when the extent is closed
 )
 
 const (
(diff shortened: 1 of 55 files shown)


aravindvs:
@venkat1109 - can you look at this rather than (https://github.com/uber/cherami-server/pull/110) ? This has all the things including the client side one. I also addressed most of the comments you had on the other PR.

aravindvs:
Note: I am trying to add the store mock tests we discussed yesterday in person..

services/inputhost/pubconnection.go
```diff
@@ -504,17 +510,61 @@ func (conn *pubConnection) flushCmdToClient(unflushedWrites int) (err error) {
 	return
 }
 
-func (conn *pubConnection) writeCmdToClient(cmd *cherami.InputHostCommand) (err error) {
+func (conn *pubConnection) writeCmdToClient(cmd *cherami.InputHostCommand, drainWG *sync.WaitGroup) (err error) {
 	if err = conn.stream.Write(cmd); err != nil {
```
venkat1109: mark the WG as done even when the write fails, that way, the caller doesn't have to await until timeout

services/inputhost/exthost.go
```diff
@@ -853,3 +870,58 @@ func (conn *extHost) getState() *admin.InputDestExtent {
...
+	if atomic.LoadInt64(&conn.seqNo) == replySeqNo {
+		ret = true
+	}
+
+	conn.lk.RUnlock()
+	return ret
+}
+
+// prepDrain just decrements the number of writable extents for this destination
+func (conn *extHost) prepDrain() bool {
```
venkat1109: Can we rename the prepDrain()/drain() and the associated states to `prepareForClose` / `closeWrite` ? drain is a bit confusing for extHost, because, it doesn't do anything special, it just prepares for close.

services/inputhost/pathCache.go
```diff
@@ -508,3 +536,26 @@ func (pathCache *inPathCache) getState() *admin.DestinationState {
 
 	return destState
 }
+
+func (pathCache *inPathCache) drainExtent(extUUID string, updateUUID string, drainWG *sync.WaitGroup) {
+	defer drainWG.Done()
+	pathCache.RLock()
+	defer pathCache.RUnlock()
```
venkat1109: is it necessary to hold the lock while you are waiting for the connDrainWG ? Imagine an input host that's been asked to drain an extent (and at the same time assigned a new extent). When this happens, the client will try to reconnect to this input host (assuming the old conn is virtually closed in response to a DRAIN). But the Open call from the client cannot succeed because the pathCache lock is being held here.

services/inputhost/pathCache.go
```diff
@@ -508,3 +536,26 @@ func (pathCache *inPathCache) getState() *admin.DestinationState {
...
+	}
+
+	if extCache.connection.prepDrain() {
+		var connDrainWG sync.WaitGroup
+		// first send DRAIN command to all connections
+		// then start draining extents
+		pathCache.drainConnections(updateUUID, &connDrainWG)
+		// do best effort waiting to notify all connections here
+		common.AwaitWaitGroup(&connDrainWG, connWGTimeout)
+		extCache.connection.drain()
```
venkat1109: remind me, why is this simply not extCache.connection.close()  ? At this point, we have waited enough for the connections to drain, all we have to do is unload everything, right ?

aravindvs: Even after we send the DRAIN message to the client and all the connections have drained (i.e, they have flushed their buffers which means only the *write* side is done), if we do a simple close here, the exthost layer (which is buffered as well) will fail any inflight messages it has in it's buffer where we could do better and get ACKs from stores as well. Also this the reason i don't use the entire drain timeout for the connection WG but rather a few seconds to give the clients a chance to flush on their side.

services/inputhost/pathCache.go
aravindvs: yeah we shouldn't hold the lock..

services/inputhost/pathCache.go
```diff
@@ -508,3 +536,29 @@ func (pathCache *inPathCache) getState() *admin.DestinationState {
...
+		// first send DRAIN command to all connections
+		// then start draining extents
+		pathCache.drainConnections(updateUUID, &connDrainWG)
+		// while drain no need to hold the lock to make sure we
+		// can proceed forward
+		pathCache.RUnlock()
+		// do best effort waiting to notify all connections here
+		common.AwaitWaitGroup(&connDrainWG, connWGTimeout)
+		extCache.connection.stopWrite()
+	}
```
aravindvs: if prepForClose() returns false, which means it the extent is already either closed or being closed, then we just leave the pathCache RLock hanging.. it must be unlocked here (pathCache.RUnlock()).. 
@GuillaumeBailey @thuningxu

PR #116, opened 2017-03-30
kafka stream implementation, import Kafka libs

GuillaumeBailey:
Added unbounded semaphore for credit signalling. There is a dropbox unbounded semaphore lib, but it takes and releases resources one at a time, which isn't what we want. This implementation is generic enough to handle size-based credits in the future.

Otherwise, this just doles out messages from the channel and counts credits


services/outputhost/kafkaStream.go
```diff
@@ -0,0 +1,119 @@
...
+
+func convertKafkaMessageToCherami(k *s.ConsumerMessage) (c *store.ReadMessageContent) {
+	c = &store.ReadMessageContent{
+		Type: store.ReadMessageContentTypePtr(store.ReadMessageContentType_MESSAGE),
+	}
+
+	c.Message = &store.ReadMessage{
+		Address: common.Int64Ptr(int64(4200000000000000000) + // arbitrary number to distinguish from offset
+			int64(1000000000000000)*int64(k.Partition) + // Make addresses for different partitions disjoint
+			k.Offset), // TODO: Topic mapping?
```
boy-uber: Suggest adding an "assert" to check offset < int64(1000000000000000) and panic if that happens. It should not happen in theory, just in case there was bug somewhere causing that, and we could fail early :)

services/outputhost/kafkaStream.go
```diff
@@ -0,0 +1,119 @@
...
+func (k *kafkaStream) Flush() error { return nil }
+
+// Done is a no-op; consumer connection is handled at a higher level
+func (k *kafkaStream) Done() error { return nil }
+
+// Read returns the next message, if:
+// * a message is available
+// * enough credit has been granted
+func (k *kafkaStream) Read() (*store.ReadMessageContent, error) {
+	k.creditSemaphore.Acquire(1)
```
boy-uber: Is it possible to add timeout mechanism here and return error on timeout. That will prevent execution stuck here due to lack of credits (e.g. code bug never sends credit).

common/unboundedSemaphore.go
```diff
@@ -0,0 +1,85 @@
...
+		if n >= s.resCount { // If we have just enough or too few resources, grab them all and maybe keep accumulating
+			n -= s.resCount
+			s.resCount = 0
+		} else { // If we have more resources than required, just subtract what we need and exit
+			s.resCount -= n
+			n = 0
+		}
+	}
+	s.acquireLock.Unlock()
+	s.c.L.Unlock()
```
KiranRG: move up one line

common/unboundedSemaphore.go
```diff
@@ -0,0 +1,85 @@
...
+		return
+	}
+
+	// Ensure that only one thread is accumulating resources at one time;
+	// Prevents acquire(1000)+acquire(1000)+release(500)+release(500) from deadlocking
+	// This may not work very well with high contention; it may take a long time to accumulate resources if there are
+	// many threads acquiring without accumulation
+	s.acquireLock.Lock()
+	for n > 0 { // While we are waiting for more resources
+		s.c.L.Lock()
```
KiranRG: move up one line

common/unboundedSemaphore.go
```diff
@@ -0,0 +1,85 @@
...
+		for s.resCount <= 0 { // Wait for resources to be available
+			s.c.Wait()
+		}
+		if n >= s.resCount { // If we have just enough or too few resources, grab them all and maybe keep accumulating
+			n -= s.resCount
+			s.resCount = 0
+		} else { // If we have more resources than required, just subtract what we need and exit
+			s.resCount -= n
+			n = 0
+		}
```
KiranRG: x = min(n, s.resCount)
n -= x
s.resCount -= x

services/outputhost/kafkaStream.go
GuillaumeBailey: I think we will just remove this entirely, since it is probably not required.

services/outputhost/kafkaStream.go
GuillaumeBailey: we have not handled closing yet, but the caller should close after read, so there will be no problem.

common/unboundedSemaphore.go
```diff
@@ -59,13 +65,14 @@ func (s *UnboundedSemaphore) Acquire(n int) {
...
 	// This may not work very well with high contention; it may take a long time to accumulate resources if there are
-	// many threads acquiring without accumulation
-	s.acquireLock.Lock()
+	// many competing threads acquiring without accumulation
+	s.accumulatorLock.Lock()
+	s.c.L.Lock()
 	for n > 0 { // While we are waiting for more resources
-		s.c.L.Lock()
 		for s.resCount <= 0 { // Wait for resources to be available
+		//	fmt.Println(`Waiting to acummulate `, n)
```
KiranRG: remove

common/unboundedSemaphore.go
```diff
@@ -24,17 +24,23 @@ import "sync"
...
+	resCount        int
+	once            sync.Once
+	c               *sync.Cond
+	m               sync.Mutex // need to name this so it's not exported. Don't use this directly, use through s.c.L
+	accumulatorLock sync.Mutex // Used by acquirer for credit accumulation
 }
 
 // Release releases n resources
 func (s *UnboundedSemaphore) Release(n int) {
 	s.once.Do(s.init)
```
KiranRG: While this is interesting in that it looks similar to a "mutex" (in not needing a New), I think it is a little weird that you have to start by dong a "Release(n)" to being using the semaphore.

I personally think it would be good to have a "New(initialCount int)" .. that takes in an "initial" count for the semaphore.

Probably not a fair comparison, since these are different platforms, nevertheless, for reference these take in an "initial" count:
java semaphore: https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Semaphore.html
win32 semaphore: https://msdn.microsoft.com/en-us/library/windows/desktop/ms682438(v=vs.85).aspx

common/unboundedSemaphore.go
```diff
@@ -24,17 +24,23 @@ import "sync"
...
+//
+// See also: https://godoc.org/github.com/dropbox/godropbox/sync2#NewUnboundedSemaphore
+//
 type UnboundedSemaphore struct {
-	resCount    int
-	once        sync.Once
-	c           *sync.Cond
-	m           sync.Mutex // Don't use this directly, use through s.c.L
-	acquireLock sync.Mutex // Used by acquirer for credit accumulation
+	resCount        int
```
KiranRG: "count" .. for terseness.

common/unboundedSemaphore.go
```diff
@@ -0,0 +1,92 @@
...
+		s.resCount--
+		s.c.L.Unlock()
+		return
+	}
+
+	// Ensure that only one thread is accumulating resources at one time;
+	// Prevents acquire(700)+acquire(300)+release(500)+release(500) from deadlocking
+	// This may not work very well with high contention; it may take a long time to accumulate resources if there are
+	// many competing threads acquiring without accumulation
+	s.accumulatorLock.Lock()
```
venkat1109: since this is primarily used for credit management, you could also consider just returning the credits that you have (as long as its > 0 and <= n) instead of blocking to acquire 'n' credits.

venkat1109: In addition, most of the credit scenarios are specific to a single connection i.e. output <-> store or output <-> client. At most, you will have one writer and one reader (one releasing and other acquiring). Also, the most common case is for the application pumps to acquire one at a time.

common/unboundedSemaphore.go
```diff
@@ -0,0 +1,92 @@
...
+// but without any limitation on the number of resources that can be released
+//
+// See also: https://godoc.org/github.com/dropbox/godropbox/sync2#NewUnboundedSemaphore
+//
+type UnboundedSemaphore struct {
+	resCount        int
+	once            sync.Once
+	c               *sync.Cond
+	m               sync.Mutex // need to name this so it's not exported. Don't use this directly, use through s.c.L
+	accumulatorLock sync.Mutex // Used by acquirer for credit accumulation
```
venkat1109: you can consider getting rid of this, see comment below.

services/outputhost/kafkaStream.go
```diff
@@ -0,0 +1,117 @@
...
+	sc "github.com/bsm/sarama-cluster"
+	"github.com/uber/cherami-server/common"
+	"github.com/uber/cherami-server/stream"
+	"github.com/uber/cherami-thrift/.generated/go/cherami"
+	"github.com/uber/cherami-thrift/.generated/go/store"
+)
+
+// kafkaStream implements the same interface and similar-enough semantics to the Cherami store interface (BStoreOpenReadStreamOutCall)
+
+type kafkaStream struct {
```
venkat1109: please add unit tests for this class.

common/unboundedSemaphore.go
venkat1109: Also, if you do the above, consider renaming your class.. the name `Semaphore` will be a bit confusing, because semaphores block until you acquire 'n'

aravindvs: @venkat1109 but the release and acquisition of credits could be contending on outputhost, if we move to a model to potentially use this to get credits from the CG cache (which is done right now) or some credit manager which keeps accumulating credits based on acks (i.e, outstanding messages), so i believe it is better to add a lock since acks come out of bands so potentially there could be multiple writers "releasing" credits as well as multiple readers (different replica connections on the same CG) "acquiring" credits..

services/outputhost/kafkaStream.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	k.creditSemaphore.Release(int(arg.GetCredits()))
+	return nil
+}
+
+// Flush is a no-op
+func (k *kafkaStream) Flush() error { return nil }
+
+// Done is a no-op; consumer connection is handled at a higher level
+func (k *kafkaStream) Done() error { return nil }
```
aravindvs: If a Read() is blocking in Acquire() and we decide to force unload the CG and hence disconnect this connection as well will the other thread will just be sitting there? Can we use the Done() to signal any waiters and unblock them with an EOF or something?

services/outputhost/kafkaStream.go
GuillaumeBailey: done

services/outputhost/kafkaStream.go
GuillaumeBailey: I switched the order around, which seems wrong, but it agrees with the future development of  size-based credits.

common/unboundedSemaphore.go
GuillaumeBailey: See Aravind's comment below.

GuillaumeBailey:
Is there anything we are waiting on for this? @aravindvs

aravindvs:
@GuillaumeBailey lgtm!

PR #118, opened 2017-03-31
controller: added support to drain extent before sealing

venkat1109:
This patch is the first step towards graceful extent sealing in controller. Specifically, this patch contains the following changes:

* On an ExtentDownEvent, the state machine is now modified to call inputHost.drain() before calling store.seal(). This ensures that an extent that's about to be sealed will first get an opportunity to drain the connected clients, before its sealed

* Added placeholder DrainExtent() implementation is input host after pulling the latest thrift changes, this can go away if #112 is merged before this PR.

What this change doesn't contain ?
* This diff doesn't contain the changes to handle graceful deployments on the input host. The proposed design for that involves input host sending its node status as part of the load reporting. This will be done on a separate PR.
* This diff also doesn't contain the changes to do store seal sequence number reconciliation

### DIFF
--- a/services/controllerhost/event_handlers.go
+++ b/services/controllerhost/event_handlers.go
@@ -824,7 +824,7 @@ func (event *ExtentDownEvent) Handle(context *Context) error {
 			event.state = drainExtentState
 
 		case drainExtentState:
-			drainExtent(context, event.dstID, event.extentID, event.inputID)
+			drainExtent(context, event.dstID, event.extentID, event.inputID, metrics.ExtentDownEventScope)
 			event.state = sealExtentState
 
 		case sealExtentState:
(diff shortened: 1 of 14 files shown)


services/controllerhost/event_handlers.go
```diff
@@ -1091,6 +1102,38 @@ func sealExtentOnStore(context *Context, storeUUID string, storeAddr string, ext
...
+		Extents:    []*admin.DrainExtents{drainExtentInfo},
+	}
+	context.log.WithFields(bark.Fields{
+		common.TagDst: common.FmtDst(dstID),
+		common.TagExt: common.FmtExt(extentID),
+		`reconfigID`:  drainReq.GetUpdateUUID(),
+	}).Info(`sending drain command to input host`)
+
+	ctx, cancel := thrift.NewContext(drainExtentTimeout)
+	adminClient.DrainExtent(ctx, drainReq)
```
aravindvs: Won't it be better to catch the error here and record a log and/or a metric for tracking?

venkat1109:
Closing this in favor of #123

PR #120, opened 2017-03-31
Add cheramiCommitter and committer interface

GuillaumeBailey:
The idea here is to separate the metadata operation and the tracking of what work is safe to commit. The Kafka version for this will be able to demux the stream of Commit() calls to the appropriate topic+partition.

Kafka Commit() will call https://godoc.org/github.com/bsm/sarama-cluster#OffsetStash.MarkPartitionOffset
Kafka Flush() will call https://godoc.org/github.com/bsm/sarama-cluster#Consumer.MarkOffsets

### DIFF
--- a/services/outputhost/ackmanager.go
+++ b/services/outputhost/ackmanager.go
@@ -26,17 +26,14 @@ import (
 	"time"
 
 	"github.com/uber-common/bark"
-	"github.com/uber/tchannel-go/thrift"
 
 	"github.com/uber/cherami-server/common"
 	"github.com/uber/cherami-server/common/metrics"
 	"github.com/uber/cherami-thrift/.generated/go/admin"
-	"github.com/uber/cherami-thrift/.generated/go/metadata"
 	"github.com/uber/cherami-thrift/.generated/go/shared"
 )
 
 const ackLevelInterval = 5 * time.Second
-const metaContextTimeout = 10 * time.Second
 const replicatorCallTimeout = 20 * time.Second
 
 type storeHostAddress int64
@@ -49,14 +46,10 @@ type (
 	}
 
 	levels struct {
-		asOf          common.UnixNanoTime   // Timestamp when this backLogLevel was calculated
-		readLevel     common.SequenceNumber // -1 = nothing received, 0 = 1 message (#0) received, etc.
-		ackLevel      common.SequenceNumber // -1 = nothing acked
-		ackLevelRate  float64               // Rate of change in ackedTo, seconds⁽⁻¹⁾
-		readLevelRate float64               // Rate of change in receivedTo, seconds⁽⁻¹⁾
-		ackLevelAddr  storeHostAddress      // Storehost address corresponding to ackLevel, -1 = nothing acked
-		readLevelAddr storeHostAddress      // Storehost address corresponding to recievedLevel, -1 = nothing read
-		lastAckedSeq  common.SequenceNumber // the latest sequence which is acked
+		asOf         common.UnixNanoTime   // Timestamp when this level was calculated
+		readLevel    common.SequenceNumber // -1 = nothing received, 0 = 1 message (#0) received, etc.
+		ackLevel     common.SequenceNumber // -1 = nothing acked
+		lastAckedSeq common.SequenceNumber // the latest sequence which is acked
 	}
 
 	// ackManager is held per CG extent and it holds the addresses that we get from the store.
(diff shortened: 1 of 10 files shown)


services/outputhost/cheramiCommitter.go
```diff
@@ -0,0 +1,133 @@
...
+	m                  sync.Mutex
+}
+
+/*
+ * Committer interface
+ */
+
+// Commit just updates the next Cherami ack level that will be flushed
+func (c *cheramiCommitter) Commit(l CommitterLevel) {
+	c.m.Lock()
```
aravindvs: As we discussed this might a bit too heavyweight since we do it for every message. We can have a locker for the committer which takes it once during some operation and drops it once it is done..

services/outputhost/cheramiCommitter.go
```diff
@@ -0,0 +1,133 @@
...
+type cheramiCommitter struct {
+	outputHostUUID     string
+	cgUUID             string
+	extUUID            string
+	connectedStoreUUID *string
+	commitLevel        CommitterLevel
+	readLevel          CommitterLevel
+	finalLevel         CommitterLevel
+	metaclient         metadata.TChanMetadataService
+	m                  sync.Mutex
```
aravindvs: Let's make this an RWMutex since we have Read and a Write..

services/outputhost/kafkaAddresser.go
```diff
@@ -0,0 +1,141 @@
...
+			`module`:    `kafkaAddresser`,
+			`topic`:     tp.Topic,
+			`partition`: tp.Partition,
+			`offset`:    offset,
+		}).Error(`Offset out of Kafka addresser design range`)
+	}
+
+	// Pegging the offset will just mean that we don't commit
+	// larger offsets to Kafka, but shouldn't cause an outage
+	offset = common.MinInt64(offset, kafkaAddresserMaxOffset)
```
aravindvs: Let's log and/or record a metric if we ever run out of this..

services/outputhost/cheramiCommitter.go
aravindvs: You can also use unsafe to do atomic store as well if we want to avoid explicit mutexes..

services/outputhost/ackmanager.go
```diff
@@ -65,12 +59,11 @@ type (
...
 		extUUID            string
 		connectedStoreUUID *string
-		*levels                    // ‡ the current levels
-		prev               *levels // ‡ the previous levels
+		*levels            // ‡ the current levels
 		ackLevelTicker     *time.Ticker
 		closeChannel       chan struct{}
 		waitConsumed       chan<- bool // waitConsumed is the channel which will signal if the extent is completely consumed given by extentCache
-		metaclient         metadata.TChanMetadataService
+		committer          Committer
```
aravindvs: Just a general thought/question.. will it make sense to move the entire ackMgr data structure within the committer? i.e, the map which we keep track of.. this way we don't need the outer lock for the ackMgr and then a separate lock for the committer and we can use the committer to do everything.. of course this means the entire logic now moves to committer..

GuillaumeBailey: The point of doing the committer was to try not to change/duplicate the Ackmanager, which has historically had difficult to find bugs. You aren't wrong, but what we'll have is basically 2 ack managers at that point.

services/outputhost/kafkaAddresser.go
GuillaumeBailey: I have a log line for this, but I'll add a metric when we do the metrics pass on all the Kafka changes.

services/outputhost/cheramiCommitter.go
GuillaumeBailey: I'm thinking that the locks in ackmanager will be sufficient. We just need to consider that the 'committer' is covered by the ackmanager lock.

services/outputhost/cheramiCommitter.go
GuillaumeBailey: I'm going to just use the ackmanager locks for protection

services/outputhost/ackmanager.go
aravindvs: hmmm.. ok.. may be we can iterate on it in the next pass.. I was thinking more of something like this but this will of course need lots of rearrangements of the code:
`Committer {
    ackMgr *StructCommonAckManager // this is the common ack manager struct between kafka and cherami but lives within the committer with locks being controlled by the committer..
    lk            sync.RWMutex
   <other necessary fields>
}`

aravindvs:
btw.. feel free to test this in one of the staging envs if you want to play with this while the review is in progress..

aravindvs:
This looks good now! Thanks for making all the changes..

I think we need an integration test as the next step in travis CI..

aravindvs:
Looks like there is some crash on the new kafka test suite..

PR #122, opened 2017-04-02
Add go vet to Travis, fix initial lints

GuillaumeBailey:


.travis.yml
```diff
@@ -67,6 +67,11 @@ install:
   - export CGO_LDFLAGS="$CGO_LDFLAGS -L`pwd` -lrocksdb"
   - export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:`pwd`"
 
+before_script:
+  - export CHERAMI_PACKAGES=`go list ./... | grep -ve vendor | grep -ve test | sed "s#github.com/uber/cherami-server#.#"`
```
venkat1109: consider moving these to a target in Makefile instead, that way, one can run this locally as well. In addition, you can reuse the #defines that already compute the set of files/packages to compile

aravindvs: agree with @venkat1109 move this to a "make lint" or something within the makefile and you can use "ALL_SRC" there to get all the files..

GuillaumeBailey: ALL_SRC won't work, since the tool wants either a list of packages, or a list of .go files in one package. I'll use something that gets reasonable coverage.

PR #125, opened 2017-04-04
Fix drain

aravindvs:
We were not actually waiting for the drain to finish before returning
but rather we just stopped the write pump (inflight messages are not yet drained).
This will mean the controller will simply seal the extent as soon
as we return.

To fix this, we need to make sure we actually wait for the drain, i.e,
the readpump is done reading all the inflight messages.

### DIFF
--- a/config/base.yaml
+++ b/config/base.yaml
@@ -90,8 +90,8 @@ ReplicatorConfig:
 
 # Logging configuration
 logging:
-  level: info
-  stdout: true
+  level: error
+  stdout: false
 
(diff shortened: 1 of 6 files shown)


aravindvs:
Just to elaborate:

Irrespective of whether connections are notified or not, i.e, this inputhost is hosting more than one extent (which means the connections can still keep writing to this host), we need to make sure the extent we are draining is *actually* drained (i.e, readpump should be closed as well) before returning rather than just stopping the write pump.

KiranRG:
Looks good to me ..

PR #127, opened 2017-04-06
controller:graceful deployments: add support for hostGoingDown state

venkat1109:
This patch adds support for a new host state in DFDD called "GoingDown" state. Any host that's going down for a planned deployment or maintenance will be in this state. This allows controller to gracefully handover extents and reduce publish errors / duplicates. The host state itself would come from the load reporter. Changes:

* Load reporter handles new host status metric
* DFDD state machine now keeps track pf goingDown hosts
* Moved the remoteZoneExtentReplication fix up into extentMon, previously was coupled with dfdd

Prerequisite: https://github.com/uber/cherami-thrift/pull/19


services/controllerhost/dfdd.go
```diff
@@ -58,246 +62,292 @@ type (
...
 
+type dfddHostState int
+
+/*
+ * State Transitions
+ *
+ *        Unknown  -- rp.HostAddedEvent           --> UP
+ *             UP  -- loadReporter.HostGoingDown  --> GoingDown
+ *             UP  -- rp.HostRemovedEvent         --> Down
+ *      GoingDown  -- rp.HostRemovedEvent         --> Down
```
datoug: can you add comment somewhere about the difference between input and store(i.e. input will removed immediately once down, while store will have the down/forgotten transition)?

services/controllerhost/dfdd.go
```diff
@@ -58,246 +62,292 @@ type (
...
+	return dfddHost{
+		state:               state,
+		lastStateChangeTime: timeSource.Now().UnixNano(),
+	}
+}
 
-	if !dfdd.context.eventPipeline.Add(event) {
-		dfdd.context.log.WithField(common.TagEvent, event).Error("Failed to enqueue event")
+func deepCopyMap(src map[string]dfddHost) map[string]dfddHost {
+	copy := make(map[string]dfddHost, common.MaxInt(len(src), 8))
```
datoug: why common.MaxInt?

services/controllerhost/event_handlers.go
```diff
@@ -99,6 +99,16 @@ type (
 		storeIDs []string
 	}
 
+	// StoreRemoteExtentReplicatorDownEvent is triggered
+	// for an extent when the primary store host that's
+	// responsible for replicating that extent from origin
+	// zone is down
+	StoreRemoteExtentReplicatorDownEvent struct {
```
datoug: is RemoteExtentPrimaryStoreDownEvent more clear? Replicator makes me this is about the replicator role

services/controllerhost/event_handlers.go
```diff
@@ -540,119 +560,109 @@ func (event *InputHostFailedEvent) Handle(context *Context) error {
...
+	})
 
-			if foundReplacement {
-				_, err = context.mm.UpdateRemoteExtentPrimaryStore(extent.GetExtent().GetDestinationUUID(), extent.GetExtent().GetExtentUUID(), newRemoteExtentPrimaryStore)
-				if err != nil {
-					lclog.WithField(common.TagErr, err).Warn(`failed to update primary store in metadata`)
-					return errRetryable
-				}
+	if !foundReplacement {
+		lclog.WithField(common.TagErr, err).Error(`failed to find replacement for old primary store`)
```
datoug: log the extent and store as well

services/controllerhost/extentmon.go
```diff
@@ -524,6 +526,41 @@ func (monitor *extentStateMonitor) handleDestinationExtent(dstDesc *shared.Desti
...
 	}
+	
+	if common.IsRemoteZoneExtent(extent.GetOriginZone(), context.localZone) {
+		monitor.handleRemoteZoneDestinationExtent(dstDesc, isDLQ, extent)
+	}
+}
+
+// handleRemoteZoneDestinationExtent is the local handler for remote destination extent
+// this handler kicks off an event if the primary replicator for the extent is down
+func (monitor *extentStateMonitor) handleRemoteZoneDestinationExtent(dstDesc *shared.DestinationDescription, isDLQ bool, extent *metadata.DestinationExtent) {
```
datoug: remove isDLQ since it's not used

services/controllerhost/extentmon.go
```diff
@@ -524,6 +526,41 @@ func (monitor *extentStateMonitor) handleDestinationExtent(dstDesc *shared.Desti
...
+	context := monitor.context
+	// handle remote zone extent replication job failures
+	var failedStores []string
+	for _, storeID := range extent.GetStoreUUIDs() {
+		state, duration := context.failureDetector.GetHostState(common.StoreServiceName, storeID)
+		switch state {
+		case dfddHostStateUnknown:
+			failedStores = append(failedStores, storeID)
```
datoug: in which scenario will we hit this?

services/controllerhost/extentmon.go
```diff
@@ -524,6 +526,41 @@ func (monitor *extentStateMonitor) handleDestinationExtent(dstDesc *shared.Desti
...
+		case dfddHostStateDown:
+			if duration >= maxHostRestartDuration {
+				failedStores = append(failedStores, storeID)
+			}
+		}
+	}
+	if len(failedStores) > 0 {
+		stats, err := context.mm.ReadExtentStats(dstDesc.GetDestinationUUID(), extent.GetExtentUUID())
+		if err != nil {
+			return
```
datoug: log the error here?

services/controllerhost/dfdd.go
venkat1109: good idea, will do.

services/controllerhost/dfdd.go
venkat1109: hmm.. we call this method from HostAddedEvent. So chances are if we allocate it to the same size as before, the map will be reallocted when we add an entry. That's why its a maxInt. If you are asking this question, it means the intention isn't clear from this code. I will remove the maxInt and pass along the desiredNewCapacity as a param or just simply leave not optimize it.

services/controllerhost/event_handlers.go
venkat1109: yeah.. I didn't like that name either.. `RemoteExtentPrimaryStore` makes me think its a store in a remote DC.. Don't like both the names :) But I will go with yours.

services/controllerhost/extentmon.go
venkat1109: let's say a store went down and immediately after the controller restarted. And let's say, the store that went down earlier never comes back online. In that case, controller will never *discover* that host, so the host state will always be unknown.

services/controllerhost/extentmon.go
venkat1109: the metadataMgr client already emits a counter for requestCount/failures. So, we will get alerted if metadata calls are failing. And since extentMon is a background job that's run every few minutes, this will be retried anyways.

services/controllerhost/api_handlers.go
```diff
@@ -88,6 +88,13 @@ func isExtentBeingSealed(context *Context, extentID string) bool {
 	return context.extentSeals.inProgress.Contains(extentID) || context.extentSeals.failed.Contains(extentID)
 }
 
+// isInputGoingDown returns true if the specified input host
+// is going down for planned maintenance or deployment
+func isInputGoingDown(context *Context, hostID string) bool {
+	state, _ := context.failureDetector.GetHostState(common.StoreServiceName, hostID)
```
venkat1109: good catch, I will add a unit test for this. Don't think this line was covered.

services/controllerhost/event_handlers.go
venkat1109: yeah, these are already included as tags in the lclog

services/controllerhost/event_handlers.go
```diff
@@ -262,9 +259,9 @@ func NewExtentDownEvent(sealSeq int64, dstID string, extentID string) Event {
 	}
 }
 
-// NewStoreRemoteExtentReplicatorDownEvent creates and returns an StoreRemoteExtentReplicatorDownEvent
-func NewStoreRemoteExtentReplicatorDownEvent(storeID string, extentID string) Event {
-	return &StoreRemoteExtentReplicatorDownEvent{
+// NewRemoteExtentPrimaryStoreDownEvent creates and returns an StoreRemoteExtentReplicatorDownEvent
```
datoug: update StoreRemoteExtentReplicatorDownEvent to the new name

PR #128, opened 2017-04-07
GetOutputHosts for Kafka destinations

kirg:
This change modifies the GetOutputHosts call to support Kafka destinations -- by creating any necessary "phantom" extents and assigning them to available outputhosts.

I am yet to write any tests for the changes.

### DIFF
--- a/services/controllerhost/consumer.go
+++ b/services/controllerhost/consumer.go
@@ -645,8 +645,6 @@ func selectNextExtentsToConsumeKafka(
 		return nil, 0, err
 	}
 
-	nAvailable := len(dstExtents)
-
 	// all open non-dlq extents should be Kafka phantom extents
 	nKafkaOpen := common.MaxInt(0, len(cgExtents.open)-nDlqOpen)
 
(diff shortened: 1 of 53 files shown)


services/controllerhost/consumer.go
```diff
@@ -568,6 +589,144 @@ func refreshCGExtents(context *Context,
 	return addExtentsToConsumerGroup(context, dstID, cgID, cgDesc.GetIsMultiZone(), newExtents, outputHosts, m3Scope), nil
 }
 
+func refreshCGExtentsKafka(context *Context,
```
venkat1109: And instead of forking at the refreshCGExtents level, we can fork at selectNextExtentsToConsume level, which was the original way the code was written.

services/controllerhost/consumer_test.go
```diff
@@ -140,6 +204,58 @@ func (s *McpSuite) TestCGExtentSelectorHonorsCreatedTime() {
...
+			nDlq++
+		}
+	}
+
+	// ensure that the number of extents picked are within bounds
+	s.Equal(numKafkaExtentsForDstKafka, nPhantom, "Wrong number of phantom extents")
+	s.Equal(maxDlqExtentsForDstKafka, nDlq, "Wrong number of phantom extents")
+}
+
 func (s *McpSuite) TestCGExtentSelectorHonorsDlqQuota() {
```
venkat1109: please add a similar test to make sure we don't make the correct number of dlq extents available for consumption

kirg: for kafka destinations, the dlq "quota" is really a static number .. and other tests already verify that we honor that.

kirg:
queueDepth and extentMon changes for Kafka will be done in separate changes.

PR #130, opened 2017-04-07
Fail GetInputHosts call for kafka destinations

kirg:


services/controllerhost/api_handlers.go
```diff
@@ -353,6 +357,13 @@ func refreshInputHostsForDst(context *Context, dstUUID string, now int64) ([]str
 	}
 
 	var dstType = getDstType(dstDesc)
+
+	// Fail attempts to publish to Kafka destinations
+	if dstType == dstTypeKafka {
```
venkat1109: how about also storing the an empty set of ip:ports as result in the result cache ? That way, if someone mistakenly tries to publish to a kafka destination and calls getInputHosts() in a loop, we won't do a metadata.readDestination() for every call.

venkat1109: We can also consider rejecting these things at the Frontend level.

kirg: I don't really like the idea of storing an empty set of ip:ports and returning them -- since we optimizing for a very uncommon case; if at all, I'd rather update the cache to support and remember to return an "error" on the request.

And yes, we should probably reject these in the Frontend. We read in the destination there anyway.

venkat1109: >  I'd rather update the cache to support and remember to return an "error" on the request.
That's cool and even better.

> " optimizing for very uncommon case"
I am ok with punting this on v1, but our system needs to be robust against bad software/config mistakenly ddos'ing us. That actually happens more commonly than actual ddos attacks, in practice :) Just to be clear, this is not an optimization, it's about robustness; its about not browning out the metadata (on bad input). 

leaving it up to you to decide if you want to take up this now or later.

kirg: I'll do it now ..

services/controllerhost/api_handlers.go
```diff
@@ -353,6 +357,13 @@ func refreshInputHostsForDst(context *Context, dstUUID string, now int64) ([]str
...
+		context.m3Client.IncCounter(m3Scope, metrics.ControllerFailures)
+		return nil, &shared.BadRequestError{Message: "Cannot publish to Kafka destinations"}
```
venkat1109: please create an error const for this.

services/controllerhost/controllerhost_test.go
```diff
@@ -387,6 +387,18 @@ func (s *McpSuite) TestGetInputHosts() {
...
 
+func (s *McpSuite) TestGetInputHostsForKafkaDest() {
+
+	dstPath := s.generateName("/cherami/mcp-test")
+	dstDesc, err := s.createDestination(dstPath, shared.DestinationType_KAFKA)
+	s.Nil(err, "Failed to create destination")
+
+	_, err = s.mcp.GetInputHosts(nil, &c.GetInputHostsRequest{DestinationUUID: dstDesc.DestinationUUID})
+	s.NotNil(err, "GetInputHosts should fail for Kafka destination")
+	_, ok := err.(*shared.BadRequestError)
```
venkat1109: Assert the specific error type (can do this if you define the error as a const)

PR #131, opened 2017-04-07
propagate cg visibility field for multi_zone destination extent

datoug:
After a dlq extent is merged to a multi_zone destination(by calling MoveExtent), the reconciliation process in remote replicator(slow path) will detect that extent and create it in remote. This patch makes sure the cg visibility field in propagated in this case.

Note there's no fast path here. When MoveExtent is called locally, this call is NOT propagated to remote zone immediately. Ideally we should but it adds lots of complexity(basically we have to expose MoveExtent as an API call in a couple of places). In the other hand, the dlq extent is only useful after the consumer group does a active zone switch(failover) so having a larger delay(10 mins) shouldn't be a big deal.

thrift change: https://github.com/uber/cherami-thrift/pull/20

### DIFF
--- a/services/replicator/metadataReconciler.go
+++ b/services/replicator/metadataReconciler.go
@@ -618,7 +618,7 @@ func (r *metadataReconciler) reconcileCgExtentMetadata(localCgs []*shared.Consum
 	return nil
 }
 
-func (r *metadataReconciler) getAllDestExtentInRemoteZone(zone string, destUUID string) (map[string]shared.ExtentStatus, error) {
+func (r *metadataReconciler) getAllDestExtentInRemoteZone(zone string, destUUID string) (map[string]*shared.ExtentStats, error) {
 	var err error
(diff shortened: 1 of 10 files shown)


services/replicator/metadataReconciler.go
```diff
@@ -803,6 +804,12 @@ func (r *metadataReconciler) reconcileDestExtent(destUUID string, localExtents m
 					OriginZone:      common.StringPtr(remoteZone),
 				},
 			}
+
+			// if the remote extent has cg visibility set (a merged dlq extent), propagate that field too.
+			if remoteExtentStats.IsSetConsumerGroupVisibility() {
```
venkat1109: what happens, when you have a DC_LOCAL consumer group for a multi-zone destination ? In that case, won't this code try to replicate the dlq extents belonging to the DC_LOCAL consumer group ?

PR #133, opened 2017-04-09
Make sure we unlock the held lock

aravindvs:
If an extent is already drained or is not active, we leave the RLock
as is when it should be unlocked!

TODO: In the process of adding more tests

### DIFF
--- a/services/inputhost/inputhost_test.go
+++ b/services/inputhost/inputhost_test.go
@@ -1459,6 +1459,125 @@ func (s *InputHostSuite) TestInputExtHostDrain() {
 	inputHost.Shutdown()
 }
 
+func (s *InputHostSuite) TestInputExtHostDrainDuplicate() {
+	destinationPath := "foo"
+
+	// setup mockService here to use the mock admin controller client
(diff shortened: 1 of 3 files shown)


GuillaumeBailey:
Since we have no performance concern with defer in this case, let's write a few new common utilities and solve this style of problem forever.

**new Common function:**

func conditionalRUnlock(unlockNeeded *bool, l *sync.RWMutex) {
	if *unlockNeeded {
		l.RUnlock()
	}
}

**Top of function:**

locked := true
defer common.conditionalRUnlock(&locked, &pathCache.RWMutex)

**Path that needs to unlock early:**

pathCache.RUnlock()
locked = false

**Other paths:**

nothing... priceless

aravindvs:
@GuillaumeBailey agreed.. i saw your autolocker.. will modify code to use that once that is in..

aravindvs:
@GuillaumeBailey having said that I would like to get this in and hotfix our inputhosts today and then refactor code to use the autolock.. does that sound reasonable?

PR #134, opened 2017-04-09
Initial kafka integration on outputhost

GuillaumeBailey:


services/outputhost/extcache.go
```diff
@@ -306,6 +330,68 @@ func (extCache *extentCache) loadReplicaStream(startAddress int64, startSequence
...
+	cfg.Group.Member.UserData, _ = json.Marshal(meta)
+
+	// Get the notifications channel; we will just log it
+	cfg.Group.Return.Notifications = true
+
+	// Older startFroms (e.g. 0, >1 weeks back) are considered to want the oldest offset
+	// TODO: Use Sarama GetMetadata to get the list of partitions, then build the offset request
+	// to use with GetAvailableOffsets, and then "somehow" manually commit it so that sarama-cluster
+	// starts from the right place
+	if common.Now()-startFrom > common.UnixNanoTime(time.Hour*24*7) {
```
boy-uber: When the startFrom is less than 1 week, will it use lastest offset? If yes, feel a little risky here, since users may not be aware of that they are actually consuming from latest even though they specify a "start from" time. Like the TODO here. And feel we'd better default to oldest offset before we implement that TODO.

services/outputhost/extcache.go
```diff
@@ -431,3 +522,61 @@ func (extCache *extentCache) getState() *admin.OutputCgExtent {
...
+			notificationNum++
+			log = log.WithField(`notificationNum`, notificationNum)
+			if len(n.Claimed) > 0 {
+				log.WithField(`claimed`, n.Claimed).Info(`claimed partitions`)
+			}
+			if len(n.Released) > 0 {
+				log.WithField(`released`, n.Released).Info(`released partitions`)
+			}
+			if len(n.Current) > 0 {
+				log.WithField(`current`, n.Released).Info(`current partitions`)
```
boy-uber: log n.Current instead of n.Released?

services/outputhost/extcache.go
```diff
@@ -306,6 +330,68 @@ func (extCache *extentCache) loadReplicaStream(startAddress int64, startSequence
...
+		kafkaTopics,
+		cfg,
+	)
+	if err != nil {
+		extCache.logger.WithField(common.TagErr, err).Error(`couldn't make Kafka client`)
+		return nil, err
+	}
+
+	// Setup the notification logger
+	go kafkaNotificationsLogger(extCache.kafkaClient.Notifications(), extCache.logger)
```
boy-uber: Suggest passing consumer group id to this kafkaNotificationsLogger function and log the consumer group id as well. The notification itself does not contain consumer group info, thus will be hard to debug.

services/outputhost/extcache.go
```diff
@@ -306,6 +330,68 @@ func (extCache *extentCache) loadReplicaStream(startAddress int64, startSequence
...
+	// Build the Kafka client. Note that we would ideally like to have a factory for this, but the client
+	// has consumer-group-specific changes to its configuration
+	extCache.kafkaClient, err = sc.NewConsumer(
+		getKafkaBrokersForCluster(kafkaCluster),
+		groupID,
+		kafkaTopics,
+		cfg,
+	)
+	if err != nil {
+		extCache.logger.WithField(common.TagErr, err).Error(`couldn't make Kafka client`)
```
boy-uber: Log consumer group and kafka topics here as well? Also better check logs in other places and add consumer group and kafka topics when possible. It will help to narrow down issues when they happen :)

services/outputhost/extcache.go
```diff
@@ -431,3 +522,61 @@ func (extCache *extentCache) getState() *admin.OutputCgExtent {
...
+			if len(n.Current) > 0 {
+				log.WithField(`current`, n.Released).Info(`current partitions`)
+			}
+		}
+	}
+	log.Info(`Notifications channel closed`)
+}
+
+func getKafkaGroupIDForCheramiConsumerGroupName(cgName string) string {
+	s := strings.Split(cgName, `/`)
```
boy-uber: Why split the consumer group path and use the last part? It may cause name conflict for kafka consumer group?

GuillaumeBailey: kafka cannot use '/' in their consumer group names. This seems like a reasonable workaround, as long as the documentation is clear.

services/outputhost/extcache.go
GuillaumeBailey: To be compatible with our current system, we need to distinguish between people that recently created a consumer group and those that want to receive all messages. In practice, I believe I have seen only one customer that had a specific startFrom time desired. This seems like an acceptable workaround.

services/outputhost/extcache.go
GuillaumeBailey: I believe the extCache.logger already has WithField(tag.CnsmID...

services/outputhost/extcache.go
boy-uber: Does the Cherami document require the full path of consumer group (before KFC) to be unique for only the last part to be unique?

Also, I tried to use / in kafka consumer group name (e.g. /kafka/consumer/group/xxx), and there is no error. So you may be able to just use the full path for kafka, or, replace "/" with "_"?

GuillaumeBailey: if that works, then I'm OK with it, but I wonder if it will break something.

GuillaumeBailey: Followed up with the Kafka team; they discourage using '/' in consumer group names, even if it will work.

services/outputhost/extcache.go
aravindvs: move the 7 day to a const..

common/stdLoggerFromBark.go
```diff
@@ -0,0 +1,21 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package common
```
aravindvs: is this used? if it is you can leave as is, if not we can remove this..

GuillaumeBailey: done

services/outputhost/extcache.go
GuillaumeBailey: done

PR #135, opened 2017-04-09
Add autoLocker

GuillaumeBailey:


common/autoLocker.go
```diff
@@ -0,0 +1,120 @@
...
+//   Pass your sync.RWMutex as-is for a W-autoLock, and use the sync.RWMutex.RLocker() for an R-autoLock
+// 
+// Example:
+// 
+// type foo struct {
+// 	l sync.RWMutex
+// }
+// 
+// func (f *foo) someFunc() {
+// 	l := AutoLock(&f.l.RLocker())
```
aravindvs: Won't it be better to provide one instance of an "AutoLocker" and then use that in other places to (R)Lock/(R)Unlock? That way everytime we won't allocate a new autolock object when we want to take a lock..

aravindvs: Something like NewAutoLocker()?

aravindvs:
@venkat1109 has some valid concerns/questions on this:
1.  Instead of providing a safety auto locker, why not just enforce the restriction on code where you always follow a Lock with a defer Unlock and rearrange the code to handle unlocks in the middle.
2. His concern about correctness on auto lock racing - which is not the case here since you always allocate a new auto locker object, which means the lifetime of this object is restricted within the scope of this routine.

We can do a post scrum on this later today.

common/autoLocker.go
```diff
@@ -0,0 +1,120 @@
...
+	l.Lock()
+	return &autoLock{
+		locked: true,
+		l:      l,
+	}
+}
+
+func AutoLockWithFinalizer(l sync.Locker) *autoLock {
+	a := AutoLock(l)
+	runtime.SetFinalizer(a, lockFinalizer)
```
venkat1109: please don't do this. Unlike C++ destructor, there are so many `conditions apply` in a garbage collected language like Java/Golang that this pattern rarely works. Just read through the documentation here https://golang.org/pkg/runtime/#SetFinalizer, if you are not convinced.

In the past, I have debugged memory leaks (because of finalizer in java) and its not fun.

kirg:
I tend to agree with previous comments; I think this isn't very idiomatic and relying on the (laziness) of a GC to unlock adds in more uncertainty .. particularly when debugging deadlocks, sync issues, etc.

I think the best way to achieve this (which I presume is a one-liner lock/unlock) in Golang is using defer mechanism, perhaps like this:

```
func AutoLock(l sync.Locker) func() {

	l.Lock()

	return func() {
		l.Unlock()
	}
}
```
And use it like this:
```

var mu sync.RWMutex

func foo() {

    defer AutoLock(&mu)()

    // do stuff under lock, that gets unlocked when we leave the function //
}

```

common/autoLocker.go
GuillaumeBailey: the idea here is to have an object that can optionally be unlocked automatically if it leaves the scope, with the finalizer. What you are saying is reasonable if you don't need the finalizer as an insurance policy. Nothing precludes using this object in that way.

common/autoLocker.go
GuillaumeBailey: This is meant as an insurance policy only, i.e. for code that isn't critical path. It provides some assurance that a simple error will not deadlock the server, just cause performance issues (since this is performed by the GC, as you mention).

GuillaumeBailey:
Kiran's suggestion is pretty clever. I'm wondering if it can be made to work with the RW lock case, where we want to upgrade the lock.

We should also look at templatizing functions like RWLockReadAndConditionalWrite, which maintains the code readability of having a single function (+ lambdas), while ensuring that mistakes can't be made.

PR #137, opened 2017-04-11
refactor selectNextExtentsToConsume before KFC changes

kirg:
I have split up the changes in GetOutputHosts to support Kafka destinations. first step, moving the extent creation code into selectNextExtentsToConsume.

### DIFF
--- a/services/controllerhost/consumer.go
+++ b/services/controllerhost/consumer.go
@@ -363,6 +363,57 @@ func fetchClassifyOpenCGExtents(context *Context, dstUUID string, cgUUID string,
 	return
 }
 
+// listConsummableExtents returns the list of extents for the destination that
+// are in open or sealed state and can be consumed by the given CG.
+func listConsummableExtents(context *Context, dstUUID string, cgUUID string, m3Scope int) ([]*m.DestinationExtent, error) {
+
+	// get list of open/sealed extents
+	filterBy := []shared.ExtentStatus{shared.ExtentStatus_SEALED, shared.ExtentStatus_OPEN}
+	dstExtents, err := context.mm.ListDestinationExtentsByStatus(dstUUID, filterBy)
+	if err != nil {
+		context.m3Client.IncCounter(m3Scope, metrics.ControllerErrMetadataReadCounter)
+		return nil, err
+	}
+
+	var consExtents []*m.DestinationExtent
+
+	for _, ext := range dstExtents {
+
+		visibility := ext.GetConsumerGroupVisibility()
+
+		// if DLQ, ensure the extent is visible to the CG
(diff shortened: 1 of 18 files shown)


services/controllerhost/consumer.go
```diff
@@ -511,7 +510,31 @@ func selectNextExtentsToConsume(
...
 
-	return result, nAvailable, nil
+	if len(cgExtents.openHealthy) == 0 && len(result) == 0 {
+
+		nBacklog := nAvailable + len(cgExtents.open)
+		maxExtentsToConsume := maxExtentsToConsumeForDst(context, dstDesc.GetPath(), cgDesc.GetConsumerGroupName(), getDstType(dstDesc), dstDesc.GetZoneConfigs())
+
+		if nBacklog < maxExtentsToConsume {
+			// No consumable extents for this destination, create one
+			extentID, _, storehosts, e := createExtent(context, dstID, dstDesc.GetIsMultiZone(), m3Scope)
```
venkat1109: can you move this logic into a helper ? controller tries to keep method length <= 100 at max.

services/controllerhost/consumer.go
```diff
@@ -397,7 +397,7 @@ func selectNextExtentsToConsume(
 	dstDesc *shared.DestinationDescription,
 	cgDesc *shared.ConsumerGroupDescription,
 	cgExtents *cgExtentsByCategory,
-	m3Scope int) ([]*m.DestinationExtent, int, error) {
```
venkat1109: how about leaving this as is ? the additional result param helps with unit testing.

services/controllerhost/consumer.go
```diff
@@ -363,6 +363,85 @@ func fetchClassifyOpenCGExtents(context *Context, dstUUID string, cgUUID string,
 	return
 }
 
+// listConsummableExtents returns the list of extents for the destination that
+// are in open or sealed state and can be consumed by the given CG.
+func listConsummableExtents(context *Context, dstUUID string, cgUUID string,
```
venkat1109: `Extents` are always prefixed with either `dst` or `cg` for clarity. 

consider renaming the signature as follows:

func findConsumableDstExtents(context, dstUUID, cgUUID string, openCGExtents map[], consumedCGExtents)

services/controllerhost/consumer.go
```diff
@@ -363,6 +363,85 @@ func fetchClassifyOpenCGExtents(context *Context, dstUUID string, cgUUID string,
 	return
 }
 
+// listConsummableExtents returns the list of extents for the destination that
+// are in open or sealed state and can be consumed by the given CG.
```
venkat1109: also make it clear that the returned dstExtents *aren't already consumed by the CG*

PR #139, opened 2017-04-11
return empty output host lists if current zone is not active zone for multi_zone cg

datoug:
This patch adds dynamic configs to controller to control the multi_zone cg active zone switch. Together with the per cg active zone override, controller will decide which zone to deliver the message and which zone to not (by returning an empty list of output hosts).

There're two dynamic configs added:
1. failoverMode: whether we're in failover mode
2. activeZone: what's the active zone

Here're the steps to decide whether to deliver message in current zone:
1. If we're in failover mode, always trust the active zone in dynamic config
2. Otherwise, check whether the per cg active zone is specified. Use that as active zone if yes
3. Otherwise, use the active zone in dynamic config

This allows both scenarios:
1. if we're in failover mode, the active zone in dynamic config takes precedence.
2. if we're in normal mode(all zones are available), we allow cg to specify their preferred active zone (so the cg owner can test their failover scenario).

The config updater is an interface that allow us to dynamically update the config based on information from some external sources(for example, some external service that detects whether some zone is down). Currently it's not implemented.


services/controllerhost/config.go
```diff
@@ -50,10 +50,16 @@ type (
...
 
+	// ControllerDynamicConfig contains the config
+	// parameters needed for controller
 	ControllerDynamicConfig struct {
 		NumPublisherExtentsByPath      []string `name:"numPublisherExtentsByPath" default:"/=4"`
 		NumConsumerExtentsByPath       []string `name:"numConsumerExtentsByPath" default:"/=8"`
 		NumRemoteConsumerExtentsByPath []string `name:"numRemoteConsumerExtentsByPath" default:"/=4"`
+
+		// configs for multi_zone consumer group
+		ActiveZone   string `name:"activeZone"`
```
venkat1109: set the default to empty string explicitly, just to make sure that the `default` tag exist.

services/controllerhost/consumer.go
```diff
@@ -659,6 +659,28 @@ func refreshOutputHostsForConsGroup(context *Context,
 			})
 	}
 
+	if cgDesc.GetIsMultiZone() {
+		cfgObj, err := context.cfgMgr.Get(common.ControllerServiceName, "*", "*", "*")
```
venkat1109: consider moving getDynamicConfig() into its own method.

common/dconfig/configupdater.go
```diff
@@ -0,0 +1,28 @@
+package dconfig
+
+import (
+	"github.com/uber/cherami-server/common"
+)
+
+// ConfigUpdater is a daemon that can be used to update the config based on information from some external source
+type (
+	ConfigUpdater interface {
```
venkat1109: nit: ConfigUpdater is confusing IMO, given that the intention of your implementation is to detect / manage / orchestrate zone failover (based on info from external source). Would prefer a specific name, something along the lines of ZoneFailoverManager (or) ZoneFailoverDetector (or) better.

venkat1109: based on the comment above, this interface / impl might belong to common instead of dconfig.

PR #141, opened 2017-04-12
Add ability to handle planned upgrades

aravindvs:
This patch adds the ability to handle upgrades by doing the following:
* Exposes a http endpoint, which can be called to perform upgrades. For example, inputhost will use this to mark itself as going down and start draining everything
* Exposes the ability to handle signals. For example, you can have a SIGTERM handler, if needed to make sure we catch it and do as is required.


aravindvs:
closing this in favor of https://github.com/uber/cherami-server/pull/143

PR #143, opened 2017-04-12
Input upgrade

aravindvs:
This is on top of https://github.com/uber/cherami-server/pull/141

This includes the following:
* Report NodeStatus as GOING_DOWN on the controller
* Start draining everything on the inputhost

### DIFF
--- a/common/servicetypes.go
+++ b/common/servicetypes.go
@@ -114,7 +114,7 @@ type (
 		// IsLimitsEnabled is used to see if we need to enforce limits on this service
 		IsLimitsEnabled() bool
 
-		// UpgradeHandler is the hanler for the upgrade end point for this service
+		// UpgradeHandler is the handler for the upgrade end point for this service
 		UpgradeHandler(w http.ResponseWriter, r *http.Request)
 	}
 )
--- a/common/servicetypes.go
+++ b/common/servicetypes.go
@@ -113,5 +113,8 @@ type (
 
 		// IsLimitsEnabled is used to see if we need to enforce limits on this service
 		IsLimitsEnabled() bool
+
+		// UpgradeHandler is the hanler for the upgrade end point for this service
+		UpgradeHandler(w http.ResponseWriter, r *http.Request)
 	}
 )
--- a/common/util.go
+++ b/common/util.go
@@ -176,9 +176,9 @@ func CreateHyperbahnClient(ch *tchannel.Channel, bootstrapFile string) *hyperbah
 	return hClient
 }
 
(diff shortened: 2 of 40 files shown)


common/servicetypes.go
```diff
@@ -113,5 +113,8 @@ type (
 
 		// IsLimitsEnabled is used to see if we need to enforce limits on this service
 		IsLimitsEnabled() bool
+
+		// UpgradeHandler is the hanler for the upgrade end point for this service
```
venkat1109: s/hanler/handler

services/inputhost/inputhost.go
```diff
@@ -864,13 +864,26 @@ func (h *InputHost) Report(reporter common.LoadReporter) {
...
+	if intervalSecs < 1 {
+		return
+	}
+
+	numConns := h.reportHostMetric(reporter, intervalSecs)
 	h.lastLoadReportedTime = now
 
 	// Also update the metrics reporter to make sure the connection gauge is incremented
-	numConns := *(hostMetrics.NumberOfConnections)
 	h.m3Client.UpdateGauge(metrics.PubConnectionStreamScope, metrics.InputhostPubConnection, numConns)
```
venkat1109: why not move this inside reportHostMetric() ?

common/util.go
```diff
@@ -760,3 +773,18 @@ func StringSetEqual(a, b []string) bool {
 	}
 	return StringSetEqual(b, a) // Above we checked only that all A are in B; check all B in A
 }
+
+// HandleSignal handles the passed in signal and calls the appropriate callback
+func HandleSignal(sig os.Signal, hostPort string, endpoint string, timeout time.Duration, handleFunc HandleSignalFunc) {
```
venkat1109: don't see this method called anywhere in this diff.

aravindvs: @venkat1109 yeah.. this is used to hook up with the appropriate deployment frameworks.. by default i don't trap any signals.. so that our normal testing can go on.. it is in the other diff with our internal deployment framework integration..

services/inputhost/inputhost.go
aravindvs: i didn't want to do m3 call when we were going down.. that's the only reason for not moving this..

common/util.go
aravindvs: let me add a UT for this..

venkat1109: Consider renaming this to SetupSignalHandler()

aravindvs: will do..

PR #147, opened 2017-04-13
Increase the prefetch size for drain test

aravindvs:
Drain test is the only test which tries to consume 1000 messages
and since we also run multiple other tests in parallel, the test is
slow and times out. Increase the prefetch size so that we can consume
everything.


aravindvs:
i will let this run multiple times and see if this particular test consistently passes or not..

PR #148, opened 2017-04-13
Trigger ExtentDownEvent when a store is in GoingDown state

venkat1109:
This patch contains controller side changes to handle graceful deployment of stores. Specifically, the controller triggers sealing of an extent when a store is in GoingDown state (which will be case when a host is about to go down for deployment).


services/controllerhost/api_handlers.go
```diff
@@ -109,6 +109,13 @@ func isInputGoingDown(context *Context, hostID string) bool {
 	return state == dfddHostStateGoingDown
 }
 
+// isStoreGoingDown returns true if the specified store host
```
aravindvs: can we just make this isHostGoingDown(context *Context, sName, hostID string) rather than having separate ones for different roles?

services/controllerhost/api_handlers.go
```diff
@@ -143,11 +150,20 @@ func areExtentStoresHealthy(context *Context, extent *m.DestinationExtent) bool
 	}
 
 	for _, h := range storeIDs {
+
+		status := "up"
```
aravindvs: We can make this up, down and goingDown an enum as well and define it as a const the top.

aravindvs:
One general question on dfdd for stores for my enlightenment. IIUC host "goingDown" event for storehosts is the same as host "down" (removed) event right (i.e, StoreHostFailedEvent)? If that is the case, in the handleHostGoingDown() event, if service == storehost then should we just call handleHostRemovedEvent() and not worry about doing anything else there that way we don't need to create additional StoreFailed event there (i am asking because i like the code now that the failed events are created in a single place which makes the code easier to understand)?

services/controllerhost/api_handlers.go
venkat1109: there is already a enum within dfdd for this. the right thing to do is to add a String() method to that enum. Let me refactor the code a bit to make it cleaner.

services/controllerhost/api_handlers.go
venkat1109: yeah.. dfdd does just that.. I will get rid of this.

venkat1109:
Yeah, I didn't do a good job of thinking through this carefully,  I focused on keeping the changes very minimal from the existing code (specifically, the way hostFailures are detected through rpm). 
My long term plan here is to only rely on DFDD for failures and not even look at RPM. So, this code will get even simpler. As for treating StoreHostGoingDown as StoreDown, you are right conceptually, that can be done.

Let me front load some of the things I said above and cleanup this code.

aravindvs:
Looks good now. Saw your comment on the storeGoingDown event vs normal storedown event so we are going to go with a storeFailed event here, which is ok.

As long as the UTs pass, I am good with this.

PR #150, opened 2017-04-14
Add upgrade handler for storehost

aravindvs:
Storehost simply reports itself as GOING_DOWN to the controller
and just waits for the default upgrade timeout which is 10s.

By this time, the hope is that the controller would have notified
the respective input hosts to start draining.

### DIFF
--- a/services/storehost/storehost.go
+++ b/services/storehost/storehost.go
@@ -1628,11 +1628,11 @@ func (t *StoreHost) UpgradeHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 // GetNodeStatus is the current status of this host
-func (h *StoreHost) GetNodeStatus() controller.NodeStatus {
-	return h.nodeStatus.Load().(controller.NodeStatus)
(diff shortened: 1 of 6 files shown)


services/storehost/storehost.go
```diff
@@ -1596,3 +1607,32 @@ func (t *StoreHost) ListExtents(tCtx thrift.Context) (res *store.ListExtentsResu
...
+	reporter := t.loadReporter.GetReporter()
+	// report as down
+	t.reportHostMetric(reporter, 1)
+
+	// wait for upgrade timeout
+	upgradeTimer := common.NewTimer(common.DefaultUpgradeTimeout)
+	defer upgradeTimer.Stop()
+	select {
+	case <-upgradeTimer.C:
+	}
```
kirg: you don't need a select here, you could just do "<-upgradeTimer.C"; that said, instead of creating timer and stopping, etc, it would be better to do just call "time.Sleep".

PR #156, opened 2017-04-18
add metrics for zone failover mgr

datoug:


common/metrics/defs.go
```diff
@@ -119,10 +119,13 @@ var GoRuntimeMetrics = map[MetricName]MetricType{
 
 // Scope enum
 const (
+	// -- Operation scopes for ZoneFailoverManager (common) --
+	ZoneFailoverMgrScope = iota
+
```
thuningxu: It's strange you prepend to an enum instead of append to the bottom. Is there a specific reason?

datoug: no specific reason, let me move it if it's causing confusion

PR #157, opened 2017-04-19
Speed up test by reordering integration test first, then unit test

thuningxu:
As the title.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,7 +70,7 @@ install:
   - export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:`pwd`"
 
 script:
-  - EMBEDROCKSDB=0 make cover_ci
+  - EMBEDROCKSDB=0 make -j2 cover_ci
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -100,20 +100,21 @@ cherami-store-tool: $(DEPS)
 
 bins: cherami-server cherami-replicator-server cherami-cli cherami-admin cherami-replicator-tool cherami-cassandra-tool cherami-store-tool
 
-cover_profile: lint bins
-	@echo Running tests:
+cover_profile: lint $(DEPS)
```
venkat1109: I think you still want the bins as a dep instead of DEPS

.travis.yml
```diff
@@ -71,6 +71,7 @@ install:
 
 script:
   - EMBEDROCKSDB=0 make cover_ci
```
venkat1109: you can just add j2 here and remove the next line (see comment below)

PR #158, opened 2017-04-19
Add authorization abstraction in very high level

boy-uber:


common/auth_util.go
```diff
@@ -0,0 +1,37 @@
...
+
+package common
+
+import (
+	"fmt"
+	"strings"
+)
+
+const (
+	resourceURNTemplateCreateDestination = "cherami:dst:%v"
```
venkat1109: urn:cherami:dst

common/auth.go
```diff
@@ -0,0 +1,86 @@
...
+
+	// Operation is type for an user operation, e.g. CreateDestination, PublishDestination
+	Operation string
+
+	// Resource is type for a Cherami resource, e.g. destination, consumer group
+	Resource string
+
+	// AuthManager is interface to do auth
+	AuthManager interface {
+		Authenticate(ctx context.Context) (Subject, error)
```
boy-uber: Yes, will add.

common/auth_util.go
boy-uber: Ah, good catch. Let me update.

PR #159, opened 2017-04-19
Make golint work on travis, by removing reference to master

GuillaumeBailey:
52 is for testing with travis; it should cause lint failure; will adjust to 1 for checkin.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ ALL_SRC := $(shell find . -name "*.go" | grep -v -e Godeps -e vendor \
 	-e ".*/_.*" \
 	-e ".*/mocks.*")
 
-RECENTLY_TOUCHED_SRC := $(shell git log --since="52 week ago" --name-only --oneline | grep "[^ ]*\.go$$" | grep -v -e Godeps -e vendor \
+RECENTLY_TOUCHED_SRC := $(shell git log --since="1 week ago" --name-only --oneline | grep "[^ ]*\.go$$" | grep -v -e Godeps -e vendor \
 	-e ".*/\//*" \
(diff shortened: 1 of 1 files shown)


thuningxu:
Another, probably better option is to do git pre-commit hook? https://golang.org/misc/git/pre-commit

GuillaumeBailey:
Doing the pre-commit hook for linting will be really annoying, since you will have to lint all your intermediate work. It's a good idea for the fmt.

Makefile
```diff
@@ -33,7 +33,7 @@ ALL_SRC := $(shell find . -name "*.go" | grep -v -e Godeps -e vendor \
 	-e ".*/_.*" \
 	-e ".*/mocks.*")
 
-TOUCHED_SRC := $(shell git diff --name-only master | grep "\.go$$" | grep -v -e Godeps -e vendor \
+RECENTLY_TOUCHED_SRC := $(shell git log --since="1 week ago" --name-only --oneline | grep "[^ ]*\.go$$" | grep -v -e Godeps -e vendor \
```
venkat1109: why is the 1 week ago needed ?

venkat1109:
>Another, probably better option is to do git pre-commit hook? https://golang.org/misc/git/pre-commit

I second this idea of adding a pre-commit hook. Its much better handle these errors at commit time than at build time. And if you are doing one-off stuff, just run `git commit --no-verify` and that will skip the pre-commit hooks.

thuningxu:
@GuillaumeBailey Let's abandon this because it only takes a few seconds to lint-check every .go files, making this optimization not necessary.

PR #162, opened 2017-04-20
Initial Kafka output metrics

GuillaumeBailey:


glide.yaml
```diff
@@ -72,3 +71,4 @@ import:
 - package: github.com/Shopify/sarama
   version: ^1.11.0
   repo: http://github.com/Shopify/sarama
+- package: github.com/rcrowley/go-metrics
```
venkat1109: Can you give a short summary of: 
* the value this new dependency brings in (that we don't already have)
* why is that important to us
* why is this the right way, going forward, say compared to https://github.com/uber-go/tally

Unless there is a strong reason to pull in this dependency, I am not inclined to go down this road.
uber-go/tally is close to what we already have and provides the missing features on top of it.

GuillaumeBailey: The strong reason is that the Kafka client library depends on this to provide metrics, hence we don't get a choice. Either this, or no Kafka metrics.

common/metrics/go-metrics-m3.go
```diff
@@ -0,0 +1,159 @@
...
+		select {
+		case <-t.C:
+		case <-r.closeCh:
+			break exportLoop
+		}
+	}
+}
+
+func (r *goMetricsExporter) export() {
+	r.registry.Each(func(name string, i interface{}) {
```
kirg: I am not sure I understand this. So for every metric in the 'go-metrics' registry, we emit into m3? This would mean, we are expected to read in every metric, right?

Wouldn't it be better if probably we only pulled in metrics that we have listed in our metrics map that we pass in with 'New' .. that way we can ignore some metrics that we aren't interested in, etc?

common/metrics/go-metrics-m3.go
```diff
@@ -0,0 +1,159 @@
...
+// metrics system and returns the go-metrics registry that should be used
+func NewGoMetricsExporter(
+	c Client,
+	scope int,
+	metricNameToMetric map[string]int,
+	l bark.Logger,
+	closeCh <-chan struct{},
+) gometrics.Registry {
+	// Make a new registy and give it to the caller. This allows the exporters to operate independently
+	registry := gometrics.NewRegistry()
```
GuillaumeBailey: The common/metrics package thing might make sense. Let's talk in person.

common/metrics/go-metrics-m3.go
```diff
@@ -0,0 +1,159 @@
...
+		registry:            registry,
+		metricNameToMetric:  metricNameToMetric,
+		l:                   l,
+		closeCh:             closeCh,
+		metricNameToLastVal: make(map[string]int64),
+
+		// Calculate the time to provide the delayToFirstLogDump in export()
+		lastLogDump: time.Now().Add(-1 * delayBetweenLogDumps).Add(delayToFirstLogDump),
+	}
+	go r.run()
```
GuillaumeBailey: I'm trying to keep this quite simple on the calling side, i.e. just an init and it handles itself, i.e. magic. I can see the argument in following an established pattern, but that seems like it just adds a few more lines of code on the caller side, and it's speculative (i.e. it presumes that the caller would want to stop the metric reporting at some time other than the natural end of its life). Let's talk in-person.

common/metrics/go-metrics-m3.go
GuillaumeBailey: The general problem here is that any metrics solution has a pattern of local-reporting->aggregation->remote-reporting, but rcrowley metrics and m3 both want to do the aggregation step. So I have to de-aggregate from rcrowley to allow m3 to aggregate.

Regarding the Each() call, this is just saying that I get a stab at handling all metrics that Sarama reports through rcrowley/go-metrics. I don't handle all of them. The unhandled ones are periodically (but minimally) reported in the unexported metrics section. I include that unexported reporting, since there may come a time where those metrics are useful for some kind of debugging. Right now, it's a bunch of broker-specific metrics, which I fear will exceed the cardinality limits that M3 imposes on us. See https://godoc.org/github.com/Shopify/sarama

For the handled metrics, I need to de-aggregate them in various ways, and then map them to the equivalent M3 stuff.

common/metrics/go-metrics-m3.go
GuillaumeBailey: Looks like making it a package also means that I have to make a new directory, which will contain just this one file.

common/goMetricsExporter/go-metrics-m3.go
```diff
@@ -0,0 +1,123 @@
...
+	return r, registry
+}
+
+// Stop stops the exporter
+func (r *GoMetricsExporter) Stop() {
+	close(r.closeCh)
+}
+
+// Run runs the exporter. It blocks until Stop() is called, so it should probably be run in a go-routine
+func (r *GoMetricsExporter) Run() {
```
kirg: having the caller do a "go Run()" seems to unnecessarily expose internals. I would (again) suggesting having a "Start()" that does a "go r.run()" internally. That goes well will the "Stop()" you have.

PR #170, opened 2017-04-25
Bug fixes of multi_zone consumer group

datoug:
Fix a couple of bugs:
1. If we couldn't get active zone from config, default to NOT consume (otherwise we may allow consuming in both zones)
2. For GetOutputHost call, make the empty host a valid case.

### DIFF
--- a/services/controllerhost/consumer.go
+++ b/services/controllerhost/consumer.go
@@ -865,7 +865,7 @@ func shouldConsumeInZone(zone string, cgDesc *shared.ConsumerGroupDescription, d
 		return strings.EqualFold(zone, dConfig.ActiveZone)
 	}
 
-	if cgDesc.IsSetActiveZone() {
+	if len(cgDesc.GetActiveZone()) > 0 {
 		return strings.EqualFold(zone, cgDesc.GetActiveZone())
 	}
 
--- a/services/controllerhost/consumer.go
+++ b/services/controllerhost/consumer.go
@@ -764,6 +764,7 @@ func refreshOutputHostsForConsGroup(context *Context,
 	var dstType = getDstType(dstDesc)
 	var outputAddrs []string
 	var outputIDs []string
+	var consumeDisabled bool
 	var outputHosts map[string]*common.HostInfo
 
 	cgDesc, err := context.mm.ReadConsumerGroup(dstID, "", cgID, "")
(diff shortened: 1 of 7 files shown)


services/controllerhost/consumer.go
```diff
@@ -859,21 +859,21 @@ func refreshOutputHostsForConsGroup(context *Context,
...
+	if len(cgDesc.GetActiveZone()) > 0 {
 		return strings.EqualFold(zone, cgDesc.GetActiveZone())
 	}
 
 	if len(dConfig.ActiveZone) > 0 {
 		return strings.EqualFold(zone, dConfig.ActiveZone)
 	}
 
-	return true
+	return false
```
venkat1109: can you also add a log or metric when both dConfig.ActiveZone and cgDesc.GetActiveZone() are empty ? 
just want to make sure we have mechanisms to verify correctness as we roll this out initially.

services/controllerhost/controllerhost.go
```diff
@@ -385,19 +385,17 @@ func (mcp *Mcp) GetOutputHosts(ctx thrift.Context, inReq *c.GetOutputHostsReques
 	}
 
 	response := func(outputHostIDs []string, err error) (*c.GetOutputHostsResult_, error) {
-		if len(outputHostIDs) < 1 {
-			// only count as failure if our answer contains no endpoints at all
+		if err != nil {
```
venkat1109: let's revert to the same code as before:
https://github.com/uber/cherami-server/blob/f47853600f4ebb2fc74a41e3a7d3e11ab6f07031/services/controllerhost/controllerhost.go

checking for err != nil still changes the behavior.

datoug:
For the bug fix in controller, I reverted back to the original change but added another variable in the cache called consumeDisabled to distinguish the multi_zone scenario(a ok case) and the no output host case(an error case).

Also I fixed another bug in replicator. In authoritative zone, we skipped the destination and cg reconciliation, so localCgs variable is always empty, which results in no cg extent reconciliation for authoritative zone. Now I'm moving the computation of localCgs to the outer scope.

datoug:
@venkat1109 could you take another look?

services/controllerhost/consumer.go
```diff
@@ -859,21 +862,23 @@ func refreshOutputHostsForConsGroup(context *Context,
...
+		return strings.EqualFold(context.localZone, cgDesc.GetActiveZone())
 	}
 
 	if len(dConfig.ActiveZone) > 0 {
-		return strings.EqualFold(zone, dConfig.ActiveZone)
+		return strings.EqualFold(context.localZone, dConfig.ActiveZone)
 	}
 
-	return true
+	context.log.Warn(`no active zone from dynamic config !`)
```
venkat1109: would be a good idea to emit a metric here

PR #181, opened 2017-04-27
outputhost: debugs logs around ackmanager

venkat1109:
This is a patch on top of the v.18.0-rc3 release with debug logs around ackmanager.

### DIFF
--- a/services/outputhost/ackmanager.go
+++ b/services/outputhost/ackmanager.go
@@ -185,6 +185,9 @@ func (ackMgr *ackManager) addLevelOffsetImpl(offset common.SequenceNumber) {
 func (ackMgr *ackManager) stop() {
 	close(ackMgr.closeChannel)
 	ackMgr.doneWG.Wait()
+	ackMgr.logger.WithFields(bark.Fields{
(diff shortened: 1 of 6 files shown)


services/outputhost/ackmanager.go
```diff
@@ -347,6 +353,13 @@ func (ackMgr *ackManager) updateAckLevel() {
 				default:
 				}
 			}
+			ackMgr.logger.WithFields(bark.Fields{
+				`status`:           oReq.GetStatus(),
+				`ackLevelAddress`:  oReq.GetAckLevelAddress(),
+				`ackLevelSeqNo`:    oReq.GetAckLevelSeqNo(),
+				`readLevelAddress`: oReq.GetReadLevelAddress(),
+				`readLevelSeqNo`:   oReq.GetReadLevelSeqNo(),
+			}).Info("ack level update succeeded")
```
datoug: will this cause too much noise since we log this for every ack level update?

venkat1109: Yup, it will be noisy. May be instead of logging every update, I will log the ack manager state when the ackmgr is stopped.

PR #182, opened 2017-04-28
Fix level-offset in outputhost-ackmgr

kirg:


services/outputhost/ackmanager.go
```diff
@@ -124,19 +124,15 @@ func (ackMgr *ackManager) getNextAckID(address int64, sequence common.SequenceNu
 	ackMgr.readLevel++ // This means that the first ID is '1'
 	ackMgr.readLevelAddr = storeHostAddress(address)
 
-	expectedReadLevel := ackMgr.levelOffset + ackMgr.readLevel
-
 	// NOTE: sequence should be zero for timer destinations, since they usually have discontinuous sequence numbers
-	if sequence != 0 && expectedReadLevel != sequence {
-		skippedMessages := sequence - expectedReadLevel
+	if sequence != 0 && sequence != ackMgr.readLevel {
```
venkat1109: can't compare readLevel with sequence here. I think you need to keep track of a new member in the struct levels that tracks the current storeSeq corresponding to the ackLevelAddr. Initially, when we restore from an earlier checkpoint, ackLevelAddr will be initialized to cge.GetAckLevelOffset() and you need to initialze storeSeq to cge.GetAckLevelSeqNo(). Then, during every getNextAckID() call, compare the received storeSeq to the current storeSeq and log if they mismatch. If there is a mismatch, also reset your in-memory storeSeq to be in sync with the value received from the store.

i.e.

```

levels struct {
   ...
   readLevelAddr storeHostAddress
   readLevelStoreSeq SequenceNumber
} 

func newAckManager() {
 ....
 ackMgr.levels = &levels{
      readLevelAddr: storeHostAddress(cge.GetReadLevelOffset()),
      readLevelStoreSeq: cge.GetAckLevelSeqNo(),
 }
}

func getNextACKID(address, sequence) {
    .... 
   ackMgr.readLevel++ // this is the output seq number, always increasing by 1
    ackMgr.readLevelStoreSeq++ // this is current store seq no, can jump due to retention
    if sequence != ackMgr.readLevelStoreSeq {
             // log mismatch
             ackMgr.readLevelStoreSeq = sequence // reset here
     }
     ackMgr.readLevelAddr = address // always update the readLevelAddr

     // now store the mapping from outputSeq -> storeSeq
     ackMgr.addrs[ackMgr.readLevel] = &internalMsg{
              addr: storeHostAddress(address),
              seq:  sequence,
     }
}

```

services/outputhost/ackmanager.go
```diff
@@ -196,21 +176,11 @@ func (ackMgr *ackManager) start() {
...
 
-	return
-}
-
-func (ackMgr *ackManager) getCurrentAckLevelSeqNo() (seqNo common.SequenceNumber) {
-	ackMgr.lk.RLock()
-	seqNo = ackMgr.levelOffset + ackMgr.ackLevel
-	ackMgr.lk.RUnlock()
-	return
+	return ackMgr.readLevelAddr, ackMgr.readLevel
```
venkat1109: return ackMgr.readLevelAddr, ackMgr.readLevelStoreSeq // return the store seq here

services/outputhost/ackmanager.go
```diff
@@ -261,16 +231,15 @@ func (ackMgr *ackManager) updateAckLevel() {
...
-				// getCurrentAckLevelOffset needs addr[ackMgr.ackLevel], so delete the previous one if it exists
-				count++
-				delete(ackMgr.addrs, curr-1)
-			} else {
+			if !addrs.acked {
 				break
 			}
+
+			update = true
+			ackMgr.ackLevel = curr
```
venkat1109: you can also assign the following here:

```
ackMgr.ackLevelAddr =  addrs.addr
ackMgr.ackLevelStoreSeq = addrs.seq // ackLevelStoreSeq will be a new member

File: services/outputhost/ackmanager.go [hunk 4]
```diff
@@ -261,16 +231,15 @@ func (ackMgr *ackManager) updateAckLevel() {
...
+			count++
+			delete(ackMgr.addrs, curr-1)
```
venkat1109: if you do the above assignments, you can just delete(ackMgr.addrs, curr) here

services/outputhost/ackmanager.go
```diff
@@ -297,9 +266,9 @@ func (ackMgr *ackManager) updateAckLevel() {
 			ExtentUUID:         common.StringPtr(ackMgr.extUUID),
 			ConnectedStoreUUID: common.StringPtr(*ackMgr.connectedStoreUUID),
 			AckLevelAddress:    common.Int64Ptr(int64(ackMgr.ackLevelAddr)),
-			AckLevelSeqNo:      common.Int64Ptr(int64(ackMgr.levelOffset + ackMgr.ackLevel)), // levelOffset adjusts the level according to what was returned from getAddressFromTimestamp
+			AckLevelSeqNo:      common.Int64Ptr(int64(ackMgr.addrs[ackMgr.ackLevel].seq)),
```
venkat1109: if you do the above assignments, then this will be just 

AckLevelSeqNo: common.Int64Ptr(ackMgr.ackLevelStoreSeq)

services/outputhost/ackmanager.go
venkat1109: also you can get rid of lines 257-258

services/outputhost/ackmanager.go
```diff
@@ -297,9 +266,9 @@ func (ackMgr *ackManager) updateAckLevel() {
...
 			ReadLevelAddress:   common.Int64Ptr(int64(ackMgr.readLevelAddr)),
-			ReadLevelSeqNo:     common.Int64Ptr(int64(ackMgr.levelOffset + ackMgr.readLevel)),
+			ReadLevelSeqNo:     common.Int64Ptr(int64(ackMgr.addrs[ackMgr.readLevel].seq)),
```
venkat1109: same comment as above, you just need to assign ackMgr.readLevelStoreSeq here

GuillaumeBailey:
This all looks reasonable. I'd like to see how you handle Venkat's suggestions, then review again.

PR #183, opened 2017-04-28
allow UpdateConsumerGroup API to update active zone for a multi_zone cg

datoug:
as title. the patch includes both fast path and slow path(reconciliation) to update active zone for a multi_zone cg.

related thrift change: https://github.com/uber/cherami-thrift/pull/21


services/replicator/metadataReconciler.go
```diff
@@ -335,7 +335,56 @@ func (r *metadataReconciler) reconcileCg(localCgs []*shared.ConsumerGroupDescrip
...
+				updateRequest.OwnerEmail = common.StringPtr(remoteCg.GetOwnerEmail())
+				cgUpdated = true
+			}
+			if localCg.GetActiveZone() != remoteCg.GetActiveZone() {
+				updateRequest.ActiveZone = common.StringPtr(remoteCg.GetActiveZone())
+				cgUpdated = true
+			}
+
+			if cgUpdated {
+				lclLg.Info(`Found cg gets updated in remote but not in local`)
```
venkat1109: long line doesn't make sense to me, please revise. Also, can you emit a metric here ?

PR #184, opened 2017-05-01
fix/debug panic from kfc integration test

kirg:


test/integration/kfc_test.go
```diff
@@ -185,6 +197,12 @@ loop:
 			part, _ := strconv.Atoi(uc[`partition`])
 			offs, _ := strconv.Atoi(uc[`offset`])
 
+			// check and skip duplicates
+			if _, ok := recvMsgs[key]; ok {
```
boy-uber: Suggest putting a counter for duplicate here

PR #189, opened 2017-05-04
Add test hook to have empty and very short extents

GuillaumeBailey:


services/inputhost/exthost.go
```diff
@@ -191,7 +190,47 @@ var (
...
+			`TestShortExtents`)
+	}
+
+	length := common.GetRandInt64(int64(extentRolloverSeqnumMin), int64(extentRolloverSeqnumMax)) // 50% chance with the test override, 100% with no test override
+	if shortExtents > 0 {
+		if common.GetRandInt64(0, 1) == 0 { // 50% chance
+			if common.GetRandInt64(0, 1) == 0 { // 50% chance
+				length = 0 // Empty extent with 25% probability with the test override
+			}
+			length = common.GetRandInt64(1, 20) // 25% chance of very short extent with the test override
```
kirg: "20" for a max is probably too small .. I'd go with something a little over the "initial-credits" sent by outputhost to a replica; so we have extents that are consumed with one credit refresh and some with a few. I think the default initial-credits is 10,000.

kirg (approved):

PR #190, opened 2017-05-04
Outputhost replica stream: write-pump error should not close read-pump

kirg:
When the store reaches the end of the extent, it closes the stream and the underlying connection -- the outputhost is expected to drain out the connection before closing out. Unfortunately, in that time, if the outputhost write-credits-pump attempts to write out credits, it would fail since the connection is closed -- and in response to that, it triggers a stoppage of the read-pump, as well, before it drains out messages in the stream. This would cause outputhost to connect to a new replica to re-read the "tail" of the messages.

This change fixes that behaviour -- so a write-pump failure does not cause the read-pump to close. The read-pump would close when it has drained out the connection and it encounters a network error.

### DIFF
--- a/services/outputhost/replicaconnection.go
+++ b/services/outputhost/replicaconnection.go
@@ -321,6 +321,9 @@ func (conn *replicaConnection) writeCreditsPump() {
 	defer conn.waitWG.Done()
 	defer conn.call.Done()
 
+	var localCredits int32 // credits accumulated
+	var creditRefresh time.Time
+
 	sendCreditsToStore := func(credits int32) (err error) {
 
 		//conn.logger.WithField(`Credits`,cFlow.GetCredits()).Debug(`Sending credits!!`)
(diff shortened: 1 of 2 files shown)


services/outputhost/replicaconnection.go
```diff
@@ -382,20 +364,26 @@ func (conn *replicaConnection) writeCreditsPump() {
 		if numMsgsRead > creditBatchSize {
 			select {
 			case credits := <-conn.creditNotifyCh: // this is the common credit channel from redelivery cache
-				conn.utilSendCredits(credits, &numMsgsRead, &totalCreditsSent)
+				if err := sendCreditsToStore(credits); err != nil {
+					return
+				}
+				numMsgsRead -= credits
```
aravindvs: Let's move this within the sendCreditsToStore() so that we don't need to decrement this every single time. You can rename this numMsgsRead as well..

PR #195, opened 2017-05-08
Fix CheramiCommitter to allow setting the final level at zero

GuillaumeBailey:


glide.lock
```diff
@@ -133,7 +133,7 @@ imports:
   - common/websocket
   - stream
 - name: github.com/uber/cherami-thrift
-  version: be4d5fc25dbf54affd1d8be5bf49bd3fc0440330
+  version: be4d5fc25dbf54affd1d8be5bf49bd3fc0440330
```
kirg: you might want to discard this change.

aravindvs: yeah.. ignore this change..

PR #201, opened 2017-05-10
Make sure to wait for DLQ routines to exit

aravindvs:
We need to make sure we actually wait for the DLQ routines to
exit.

### DIFF
--- a/services/outputhost/deadletterqueue.go
+++ b/services/outputhost/deadletterqueue.go
@@ -82,7 +82,7 @@ type deadLetterQueue struct {
 	closeCh           chan struct{}
 	consumerM3Client  metrics.Client
 	m3Client          metrics.Client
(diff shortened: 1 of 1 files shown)


services/outputhost/deadletterqueue.go
```diff
@@ -81,6 +82,7 @@ type deadLetterQueue struct {
 	closeCh           chan struct{}
 	consumerM3Client  metrics.Client
 	m3Client          metrics.Client
+	dlqWG             sync.WaitGroup
```
kirg: you might want to call it just 'wg' .. when we use it, we do "dlq.dlqWg.Wait()", etc which has "stutter" (https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters).

PR #203, opened 2017-05-11
Kafka DLQ fixes & CLI improvement

GuillaumeBailey:


tools/common/lib.go
```diff
@@ -1338,11 +1347,34 @@ func Consume(c *cli.Context, cClient ccli.Client) {
...
+				Offset:      o,
+			})
+			if err != nil {
+				panic(err)
+			}
+			if err == nil {
+				_, err = fmt.Fprintf(os.Stdout, "%s\n", string(b))
+			}
+		} else {
+			headerOnce.Do(func() {
```
kirg: Why does this need to be synchronized?

services/outputhost/deadletterqueue.go
```diff
@@ -204,6 +204,7 @@ func (dlq *deadLetterQueue) publish() {
 				err = backoff.Retry(func() error {
 					receipt := dlq.publisher.Publish(&client.PublisherMessage{
 						Data: msg.GetPayload().GetData(),
+						UserContext: msg.GetPayload().GetUserContext(),
 					})
```
kirg: there seems to be a lint failure on this file ..

tools/common/lib.go
```diff
@@ -1298,6 +1298,15 @@ func Publish(c *cli.Context, cClient ccli.Client) {
 	close(receiptCh)
 }
 
+type kafkaMessageJSON struct {
+	EnqueueTime string
+	Key         string
```
boy-uber: Kafka message key and Value could be binary. Is there any specific reason that they are strings here?

tools/common/lib.go
GuillaumeBailey: We want to print that header line once, and only if we are outputting non-kafka messages. This seems like it's more idiomatic than a simple boolean, for doing the header row exactly once.

tools/common/lib.go
GuillaumeBailey: They must be string, or they will be outputted in Base64. I had a version before with []byte, and it is not very useful because of this.

PR #205, opened 2017-05-16
Include destination path in CreateDestination resource URN

boy-uber:


common/auth_util.go
```diff
@@ -26,12 +26,35 @@ import (
...
+	return fmt.Sprintf(resourceURNTemplateCreateDestination, strings.ToLower(deploymentName), strings.ToLower(dstPathString))
+}
+
+func getPathRootName(path *string) string {
+	if path == nil {
+		return ""
+	}
+
+	parts := strings.Split(*path, "/")
+	if len(parts) == 0 {
```
datoug: I don't think this condition will ever be satisfied. Even for empty string, len(parts) will be 1

boy-uber: You are right :) Let me change the code a little bit.

PR #207, opened 2017-05-18
Delete some useless code

thuningxu:
`readMessagesPump()` has some code to compute `readBatchSize` which is never used later.

### DIFF
--- a/services/outputhost/replicaconnection.go
+++ b/services/outputhost/replicaconnection.go
@@ -185,15 +185,6 @@ func (conn *replicaConnection) readMessagesPump() {
 	// lastSeqNum is used to track whether our sequence numbers are
 	// monotonically increasing
 	var lastSeqNum = int64(conn.startingSequence)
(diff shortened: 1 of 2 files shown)


services/outputhost/replicaconnection.go
```diff
@@ -185,15 +176,6 @@ func (conn *replicaConnection) readMessagesPump() {
...
-	// this should be less than or equal to the creditBatchSize because we
-	// need to make sure we are listening for credits on the other pump properly
-	readBatchSize := minReadBatchSize
-
-	if readBatchSize < int(conn.initialCredits/10) {
-		readBatchSize = int(conn.initialCredits / 10)
-	}
 
 	for {
 		select {
```
datoug: This will change the behavior. Previously we'll drain all messages before returning even if close channel is set. The new behavior will not drain messages.

thuningxu: You are right.

PR #213, opened 2017-05-24
Add auth resource to CreateConsumerGroup and refactor code a little bit

boy-uber:


services/frontendhost/frontend.go
```diff
@@ -1110,17 +1101,10 @@ func (h *Frontend) CreateConsumerGroup(ctx thrift.Context, createRequest *c.Crea
...
-	authSubject, err := h.GetAuthManager().Authenticate(ctx)
-	if err != nil {
-		// TODO add metrics
-		return nil, err
-	}
-
 	authResource := common.GetResourceURNCreateConsumerGroup(h.SCommon, createRequest.DestinationPath)
-	err = h.GetAuthManager().Authorize(authSubject, common.OperationRead, common.Resource(authResource))
+
+	err = h.checkAuth(ctx, authResource, common.OperationRead, lclLg)
```
datoug: should OperationRead actually be OperationCreate?

boy-uber: Here we use "Read Destination" as the permission to create consumer group. Try to model the CRUD permission on destination.

PR #216, opened 2017-05-31
Fix path-regex to allow for number only strings

kirg:
Our internal canaries use a shortened-uuid to in the latter part of the path to create destinations/CGs. Occasionally, when this results in a pure numeric string, the create destination/cg fails .. because our regular expressions that validate paths seem to (incorrectly?) enforce there is at least one 'alphabetic' character in each part of the path -- while our comment seems to indicate that we are only trying to enforce each part has at least one non-symbolic character.

### DIFF
--- a/common/resolver.go
+++ b/common/resolver.go
@@ -127,13 +127,13 @@ func (r *resolverImpl) cachePut(key string, value string) {
 // so names like /./. aren't valid
 
 // PathRegex regex for destination path
-var PathRegex = regexp.MustCompile(`^/[\w.]*[a-zA-Z][\w.]*/[\w.]*[[:alnum:]][\w.]*$`)
+var PathRegex = regexp.MustCompile(`^/[\w.]*[[:alnum:]][\w.]*/[\w.]*[[:alnum:]][\w.]*$`)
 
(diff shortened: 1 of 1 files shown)


common/resolver.go
```diff
@@ -127,13 +127,13 @@ func (r *resolverImpl) cachePut(key string, value string) {
 // so names like /./. aren't valid
 
 // PathRegex regex for destination path
-var PathRegex = regexp.MustCompile(`^/[\w.]*[a-zA-Z][\w.]*/[\w.]*[a-zA-Z][\w.]*$`)
+var PathRegex = regexp.MustCompile(`^/[\w.]*[[:alnum:]][\w.]*/[\w.]*[[:alnum:]][\w.]*$`)
```
GuillaumeBailey: Do you have pure numeric path components? /w already allows a-z, A-Z, 0-9, including the _ (underscore), so it seems like you can have a bunch of numbers as long as you have at least one letter with the current system.

kirg: Yes, the canary uses a 'shortened' UUID for the second part of the component .. and that sometimes results in a purely numeric string, which would fail to match the origin regex (that expects at least one non-numeric character).

PR #218, opened 2017-06-02
support updating zone configs for multi-zone dest and cg

datoug:
This patch adds support for upgrading a regular destination or consumer group to a multi-zone one.
related thrift change in: https://github.com/uber/cherami-thrift/pull/23

### DIFF
--- a/clients/metadata/metadata_cassandra.go
+++ b/clients/metadata/metadata_cassandra.go
@@ -1533,6 +1533,13 @@ func updateCGDescIfChanged(req *shared.UpdateConsumerGroupRequest, cgDesc *share
 		isChanged = true
 		cgDesc.ActiveZone = common.StringPtr(req.GetActiveZone())
 	}
+
+	if req.IsSetZoneConfigs() && !common.AreCgZoneConfigsEqual(cgDesc.GetZoneConfigs(), req.GetZoneConfigs()) {
+		isChanged = true
+		cgDesc.ZoneConfigs = req.GetZoneConfigs()
+		cgDesc.IsMultiZone = common.BoolPtr(true)
+	}
+
 	return isChanged
 }
 
--- a/clients/metadata/metadata_cassandra.go
+++ b/clients/metadata/metadata_cassandra.go
@@ -726,6 +726,12 @@ func (s *CassandraMetadataService) UpdateDestination(ctx thrift.Context, updateR
 	if !updateRequest.IsSetChecksumOption() {
 		updateRequest.ChecksumOption = common.InternalChecksumOptionPtr(existing.GetChecksumOption())
 	}
+	isMultiZone := existing.GetIsMultiZone()
+	if !updateRequest.IsSetZoneConfigs() {
+		updateRequest.ZoneConfigs = existing.GetZoneConfigs()
+	} else {
+		isMultiZone = true
+	}
 	batch := s.session.NewBatch(gocql.LoggedBatch) // Consider switching to unlogged
 
 	batch.Query(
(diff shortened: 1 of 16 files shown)


common/util.go
```diff
@@ -803,3 +804,54 @@ func AreKafkaPhantomStores(storeUUIDs []string) bool {
...
+// AreDestinationZoneConfigsEqual determines whether two zone configs have the same content
+func AreDestinationZoneConfigsEqual(left []*shared.DestinationZoneConfig, right []*shared.DestinationZoneConfig) bool {
+	if len(left) != len(right) {
+		return false
+	}
+	for _, l := range left {
+		zone := l.GetZone()
+		foundMatch := false
+		for _, r := range right {
+			if r.GetZone() == zone {
```
kirg: btw, another way to do this would obviously be to do run through one of the slices and put it into a map; and then run through the other while looking up against the map. but i guess the zone-config list would be really small to make a difference between O(n^2) vs O(n)?

services/controllerhost/controllerhost.go
```diff
@@ -1251,3 +1279,97 @@ func (mcp *Mcp) CreateRemoteZoneConsumerGroupExtent(ctx thrift.Context, createRe
 	lclLg.Info("CreateRemoteZoneConsumerGroupExtent: Extent added to consumer group")
 	return nil
 }
+
+// ValidateDestZoneConfigUpdateRequest validates the zone configs for a UpdateDestinationRequest
+func (mcp *Mcp) ValidateDestZoneConfigUpdateRequest(ctx thrift.Context, logger bark.Logger, updateRequest *shared.UpdateDestinationRequest) (bool, error) {
```
kirg: I presume you want to name this to not have "update" in it's name .. since you will be probably calling the exact same function from the CreateDestination path too.

services/controllerhost/controllerhost.go
```diff
@@ -1251,3 +1279,97 @@ func (mcp *Mcp) CreateRemoteZoneConsumerGroupExtent(ctx thrift.Context, createRe
...
+		} else {
+			// some other error
+			return false, err
+		}
+	}
+	return true, nil
+}
+
+// ValidateCgZoneConfigUpdateRequest validates the zone configs for a UpdateConsumerGroupRequest
+func (mcp *Mcp) ValidateCgZoneConfigUpdateRequest(ctx thrift.Context, logger bark.Logger, updateRequest *shared.UpdateConsumerGroupRequest) (bool, error) {
```
kirg: as with 'ValidateDestZoneConfigUpdateRequest', this could get called from the CreateDestination path too, right?

clients/metadata/metadata_cassandra.go
```diff
@@ -726,6 +726,12 @@ func (s *CassandraMetadataService) UpdateDestination(ctx thrift.Context, updateR
 	if !updateRequest.IsSetChecksumOption() {
 		updateRequest.ChecksumOption = common.InternalChecksumOptionPtr(existing.GetChecksumOption())
 	}
+	isMultiZone := existing.GetIsMultiZone()
+	if !updateRequest.IsSetZoneConfigs() {
+		updateRequest.ZoneConfigs = existing.GetZoneConfigs()
+	} else {
+		isMultiZone = true
```
datoug: That's right, we disallow single zone config in CLI part. Zone config(s) are designed for multi-zone only.

services/controllerhost/controllerhost.go
```diff
@@ -1251,3 +1279,97 @@ func (mcp *Mcp) CreateRemoteZoneConsumerGroupExtent(ctx thrift.Context, createRe
...
+			Zone: common.StringPtr(zoneConfig.GetZone()),
+			Request: &shared.ReadDestinationRequest{
+				Path: common.StringPtr(destPath),
+			},
+		})
+		if err == nil {
+			// path exists in remote. If uuid is different, then fail the validation
+			if remoteDest.GetDestinationUUID() != updateRequest.GetDestinationUUID() {
+				logger.WithField(common.TagZoneName, common.FmtZoneName(zoneConfig.GetZone())).Error(`destination exists in remote but UUID is different`)
+				return false, nil
```
datoug: actually in caller we check if there's any error. If not (and validation fails), we return BadRequestError

kirg: i do see that .. but this is bug-prone. typically/canonically, the code should always check the error-code and only if it is non-nil look at the 'value' returned.

PR #224, opened 2017-06-14
Cli bug fix: strip the password in destination path

datoug:
For the hacky security check, read destination doesn't handle that, so strip the password before issuing the read destination call

### DIFF
--- a/tools/common/lib.go
+++ b/tools/common/lib.go
@@ -327,7 +327,7 @@ func UpdateDestination(c *cli.Context, cClient ccli.Client, mClient mcli.Client,
 	// issue with cross zone replication(for example, network down between zones)
 	existingDesc, err := readDestinationFromMetadata(mClient, path)
 	ExitIfError(err)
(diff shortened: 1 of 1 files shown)


tools/common/lib.go
```diff
@@ -810,16 +804,24 @@ func ReadDestination(c *cli.Context, mClient mcli.Client) {
 	}
 }
 
+// path passed from tools might contain a password for some hacky security check, however the ReadDestination
+// metadata call doesn't handle that. This helper function strips the password if there's any.
+// TODO: we should remove this once the hacky security check is removed
+func readDestinationFromMetadata(mClient mcli.Client, path string) (*shared.DestinationDescription, error) {
+	parts := strings.Split(path, `+`)
+	return mClient.ReadDestination(&shared.ReadDestinationRequest{
+		Path: &parts[0],
```
kirg: do it inline?
```
Path: common.StringPtr(strings.Split(path, `+`)[0]),
```

PR #230, opened 2017-06-20
Handle delay and skip-older in outputhost

kirg:


services/outputhost/replicaconnection.go
```diff
@@ -186,124 +188,176 @@ func (conn *replicaConnection) readMessagesPump() {
...
-				// Wait until the there are some listeners or we are shutting down
+			if conn.extCache.destType != shared.DestinationType_TIMER {
+
+				enqueueTime := common.UnixNanoTime(msg.Message.GetEnqueueTimeUtc())
+				now := common.Now()
+
+				// check if the messages have already 'expired'; ie, if it falls outside the
+				// skip-older window. ignore (ie, don't skip any messages), if skip-older is '0'.
+				if skipOlderNanos > 0 && enqueueTime < (now-skipOlderNanos) {
+					continue loop
```
datoug: I think we should call updateSuccessfulSendToMsgsCh for issuing more credits?

services/outputhost/replicaconnection.go
```diff
@@ -186,124 +188,176 @@ func (conn *replicaConnection) readMessagesPump() {
...
+				}
+
+				// check if this is a delayed-cg, and if so delay messages appropriately
+				if delayNanos > 0 {
+
+					// compute visibility time based on the enqueue-time and specified delay
+					if visibilityTime := enqueueTime + delayNanos; visibilityTime > now {
+
+						// ref: https://golang.org/pkg/time/#Timer.Reset
+						if !delayTimer.Stop() {
```
datoug: you can use common.timer so you don't need this

datoug:
unit test or integration test?

services/outputhost/replicaconnection.go
kirg: no, we should not have to. since we have dropped the message on the floor (ie, not sent it out) we should not increment 'localMsgs' and therefore not have to renew credits, etc.

services/outputhost/replicaconnection.go
kirg: will use that.

kirg:
Unit + integration tests pending.

services/outputhost/replicaconnection.go
```diff
@@ -186,124 +188,171 @@ func (conn *replicaConnection) readMessagesPump() {
...
 
-				cMsg.AckId = common.StringPtr(conn.extCache.ackMgr.getNextAckID(storeHostAddress(msg.GetAddress()), msgSeqNum))
-				// write the message to the msgsCh so that it can be delivered
-				// after being stored on the cache.
-				// 1. either there are no listeners
-				// 2. the buffer is full
-				// Wait until the there are some listeners or we are shutting down
+			if conn.extCache.destType != shared.DestinationType_TIMER {
+
+				enqueueTime := common.UnixNanoTime(msg.Message.GetEnqueueTimeUtc())
```
GuillaumeBailey: Need to handle zero enqueue time, which is common with Kafka

services/outputhost/replicaconnection.go
```diff
@@ -186,124 +188,171 @@ func (conn *replicaConnection) readMessagesPump() {
 	// monotonically increasing
 	var lastSeqNum = int64(conn.startingSequence)
 
+	var skipOlderNanos = common.UnixNanoTime(conn.extCache.skipOlder * int32(time.Second))
```
GuillaumeBailey: skipOlder should also respect the delay time, as discussed earlier. Just need to add the delay here.

kirg:
dropping this PR .. due to merge mess.

PR #236, opened 2017-06-26
Add input validation for cli tool

boy-uber:


tools/common/lib.go
```diff
@@ -73,27 +73,57 @@ const (
 	ConsumerGroupType = "CG"
 	// MinUnconsumedMessagesRetentionForMultiZoneDest is the minimum unconsumed retention allowed
 	MinUnconsumedMessagesRetentionForMultiZoneDest = 3 * 24 * 3600
+	// MinConsumedMessagesRetention is the minimum consumed retention
+	MinConsumedMessagesRetention = 60
+	// MaxConsumedMessagesRetention is the maximum consumed retention
+	MaxConsumedMessagesRetention = 100 * 24 * 3600
```
datoug: I think 100 days is too long, maybe change to 30 days?

datoug: same for unconsumed retention

tools/common/lib.go
```diff
@@ -73,27 +73,57 @@ const (
...
+	// MinUnconsumedMessagesRetention is the minimum unconsumed retention
+	MinUnconsumedMessagesRetention = 60
+	// MaxUnconsumedMessagesRetention is the maximum unconsumed retention
+	MaxUnconsumedMessagesRetention = 100 * 24 * 3600
+	// MinLockTimeoutSeconds is the minimum lock timeout seconds
+	MinLockTimeoutSeconds = 0
```
datoug: too short? maybe 5 seconds

tools/common/lib.go
```diff
@@ -73,27 +73,57 @@ const (
...
+	// MaxLockTimeoutSeconds is the maximum lock timeout seconds
+	MaxLockTimeoutSeconds = 100 * 24 * 3600
```
datoug: This would be too long. I think generally lock timeout is in seconds. Maybe 1hr is more than enough

tools/common/lib.go
```diff
@@ -73,27 +73,57 @@ const (
...
+	// MinMaxDeliveryCount is the minimum value for max delivery count
+	MinMaxDeliveryCount = 0
```
datoug: min is 1

tools/common/lib.go
```diff
@@ -73,27 +73,57 @@ const (
...
+	// MaxMaxDeliveryCount is the maximum value for max delivery count
+	MaxMaxDeliveryCount = math.MaxInt32
```
datoug: we need a cap. Maybe look at all cgs on prod, and find out what's the max we're using today

tools/common/lib.go
```diff
@@ -73,27 +73,57 @@ const (
...
+	// MinConsumedMessagesRetention is the minimum consumed retention
+	MinConsumedMessagesRetention = 180
+	// MaxConsumedMessagesRetention is the maximum consumed retention
+	MaxConsumedMessagesRetention = 7 * 24 * 3600
+	// MinUnconsumedMessagesRetention is the minimum unconsumed retention
+	MinUnconsumedMessagesRetention = 180
+	// MaxUnconsumedMessagesRetention is the maximum unconsumed retention
+	MaxUnconsumedMessagesRetention = 7 * 24 * 3600
+	// MinLockTimeoutSeconds is the minimum lock timeout seconds
+	MinLockTimeoutSeconds = 0
```
datoug: I think this should be at least 10 seconds

tools/common/lib.go
```diff
@@ -73,27 +73,57 @@ const (
...
+	// MaxLockTimeoutSeconds is the maximum lock timeout seconds
+	MaxLockTimeoutSeconds = 3600
+	// MinMaxDeliveryCount is the minimum value for max delivery count
+	MinMaxDeliveryCount = 0
```
datoug: min 1?

boy-uber: yes, just see the comment :)

tools/common/lib.go
boy-uber: Sounds good. Let me update.

PR #237, opened 2017-06-28
Add UpdateHostsForDeployment API to replicatorClientFactory so that replicator host is updatable

datoug:


jprojs:
You should implement zookeeper server for host updater

common/configure/commonreplicatorconfig.go
```diff
@@ -42,3 +47,14 @@ func (r *ReplicatorConfig) GetReplicatorHosts() map[string]string {
 func (r *ReplicatorConfig) GetDefaultAuthoritativeZone() string {
 	return r.DefaultAuthoritativeZone
 }
+
+// GetUseStandalone checks whether a specific deployment is using standalone deployment
+func (r *ReplicatorConfig) GetUseStandalone(deployment string) bool {
```
thuningxu: I feel this is irrelevant to open source users.

cmd/servicecmd/servicestartcmd.go
```diff
@@ -254,7 +254,14 @@ func StartReplicatorService() {
 	dClient := dconfigclient.NewDconfigClient(cfg.GetServiceConfig(serviceName), serviceName)
 	sCommon := common.NewService(serviceName, uuid.New(), cfg.GetServiceConfig(serviceName), common.NewUUIDResolver(meta), hwInfoReader, reporter, dClient, common.NewBypassAuthManager())
 
-	h, tc := replicator.NewReplicator(serviceName, sCommon, meta, replicator.NewReplicatorClientFactory(cfg, common.GetDefaultLogger()), cfg)
+	allHosts := cfg.GetReplicatorConfig().GetReplicatorHosts()
+	allSplitHosts := make(map[string][]string)
+	for deployment, hosts := range allHosts {
```
thuningxu: I almost feel `GetReplicatorHosts()` should include this logic and return `map[string][]string`. Thoughts?

datoug: These logic is more like a util function rather than some logic that is specific to replicator, so I slightly feel better to keep it outside.

common/configure/commonreplicatorconfig.go
datoug: Yea, open source is not using this config. Right now we don't have a clean way to have 'internal only' config(interface is defined in open source) so keep it here for now.

PR #240, opened 2017-06-30
replicator fd leak: close the server stream on failure cases so that the tcp connection can be actually closed

datoug:
as title. make sure the server calls close on failure cases.

### DIFF
--- a/services/replicator/replicator.go
+++ b/services/replicator/replicator.go
@@ -219,6 +219,10 @@ func (r *Replicator) OpenReplicationReadStreamHandler(w http.ResponseWriter, req
 		}).Error("Can't open store host read stream")
 		r.m3Client.IncCounter(metrics.OpenReplicationReadScope, metrics.ReplicatorCreateOutStreamFailure)
 		r.m3Client.IncCounter(metrics.OpenReplicationReadScope, metrics.ReplicatorFailures)
+
+		// Must close the connection on server side because closing on client side doesn't actually close the
+		// underlying TCP connection
+		inStream.Done()
 		return
 	}
 	outConn := newOutConnection(extUUID, outStream, r.logger, r.m3Client, metrics.OpenReplicationReadScope)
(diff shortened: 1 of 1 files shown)


services/replicator/replicator.go
```diff
@@ -219,6 +219,10 @@ func (r *Replicator) OpenReplicationReadStreamHandler(w http.ResponseWriter, req
 		}).Error("Can't open store host read stream")
 		r.m3Client.IncCounter(metrics.OpenReplicationReadScope, metrics.ReplicatorCreateOutStreamFailure)
 		r.m3Client.IncCounter(metrics.OpenReplicationReadScope, metrics.ReplicatorFailures)
+
+		// Must close the connection on server side because closing on client side doesn't actually close the
+		// underlying TCP connection
+		inStream.Done()
```
kirg: Probably better do a "defer inStream.Done()" immediately after the open and if-error block.

services/replicator/replicator.go
```diff
@@ -276,6 +280,10 @@ func (r *Replicator) OpenReplicationRemoteReadStreamHandler(w http.ResponseWrite
 		}).Error("Can't open remote replication read stream")
 		r.m3Client.IncCounter(metrics.OpenReplicationRemoteReadScope, metrics.ReplicatorCreateOutStreamFailure)
 		r.m3Client.IncCounter(metrics.OpenReplicationRemoteReadScope, metrics.ReplicatorFailures)
+
+		// Must close the connection on server side because closing on client side doesn't actually close the
+		// underlying TCP connection
+		inStream.Done()
```
kirg: defer inStream.Done() .. like above.

services/replicator/replicator.go
datoug: problem is in regular cases stream is closed in the inconnection goroutine. If we do a defer, two goroutines might try to close the stream. From the websocket API the close functions are not thread-safe, also I'm not sure about the behavior of double close.

kirg: yes, close isn't thread-safe. i see .. i hadn't noticed that; this is fine then.

PR #246, opened 2017-07-06
CLI: fix defaults; make consistent with admin cli

kirg:


cmd/tools/admin/main.go
```diff
@@ -22,10 +22,22 @@ package main
 
 import (
 	"os"
+	"time"
 
 	"github.com/codegangsta/cli"
 	"github.com/uber/cherami-server/common"
 	"github.com/uber/cherami-server/tools/admin"
+	com "github.com/uber/cherami-server/tools/common"
```
datoug: rename to toolscommon for better readability

datoug:
We probably should just deprecate these commands, which are duplicates from cherami-cli. I don't see the value of having the same commands in two tools, and they're causing code duplicates like this.

Or as @thuningxu has mentioned before, we can fold these two tools into a single binary, and only expose the 'admin' functionalities to admins. Although I personally slightly prefer the current solution.

cmd/tools/cli/main.go
```diff
@@ -123,7 +122,7 @@ func main() {
 						},
 						cli.StringSliceFlag{
 							Name:  "zone_config, zc",
-							Usage: "Zone configs for multi_zone destinations. Format for each zone should be \"ZoneName,AllowPublish,AllowConsume,ReplicaCount\". For example: \"zone1,true,true,3\"",
+							Usage: "Zone configs for multi_zone destinations. Format for each zone should be \"Zone,AllowPublish,AllowConsume,ReplicaCount\". Ex: \"sjc1a,true,true,3\"",
```
datoug: we don't want to use 'sjc1' in the open source repo

cmd/tools/cli/main.go
```diff
@@ -27,18 +27,17 @@ import (
...
-
-	strDelaySeconds = `Delay to introduce to all messages, in seconds.`
-	intDelaySeconds = 0 // zero delay, by default
+	usageCGStartTime                  = `Consume messages newer than this time in unix-nanos (default: Now; ie, consume no previously published messages)`
+	usageCGLockTimeoutSeconds         = `Acknowledgement timeout for prefetched/received messages`
+	usageCGMaxDeliveryCount           = `Max number of times a message is delivered before it is sent to the DLQ (dead-letter queue)`
+	usageCGSkipOlderMessagesInSeconds = `Skip messages older than this duration, in seconds ('0' to skip none)`
+	usageCGDelaySeconds               = `Delay, in seconds, to defer all messages by`
+	usageCGOwnerEmail                 = "Owner email"
+	usageCGZoneConfig                 = "Zone configs for multi-zone CG. For each zone, specify \"Zone,PreferedActiveZone\"; ex: \"dca1a,false\""
```
datoug: no 'dca1' or 'sjc1' in the open source repo

cmd/tools/admin/main.go
```diff
@@ -22,10 +22,22 @@ package main
...
+)
+
+const (
+	usageCGStartTime                  = `Consume messages newer than this time in unix-nanos (default: Now; ie, consume no previously published messages)`
+	usageCGLockTimeoutSeconds         = `Acknowledgement timeout for prefetched/received messages`
+	usageCGMaxDeliveryCount           = `Max number of times a message is delivered before it is sent to the DLQ (dead-letter queue)`
+	usageCGSkipOlderMessagesInSeconds = `Skip messages older than this duration, in seconds ('0' to skip none)`
+	usageCGDelaySeconds               = `Delay, in seconds, to defer all messages by`
+	usageCGOwnerEmail                 = "Owner email"
+	usageCGZoneConfig                 = "Zone configs for multi-zone CG. For each zone, specify \"Zone,PreferedActiveZone\"; ex: \"dca1a,false\""
```
datoug: no dca1

tools/common/lib.go
```diff
@@ -71,6 +71,10 @@ const (
 	DestinationType = "DST"
 	// ConsumerGroupType is the name for entity type for consumer group in listEntityOps
 	ConsumerGroupType = "CG"
+	// DefaultUnconsumedMessagesRetention is the default value for unconsumed messages retention
+	DefaultUnconsumedMessagesRetention = 7200
+	// DefaultConsumedMessagesRetention is the default value for consumed messages retention
+	DefaultConsumedMessagesRetention = 3600
```
datoug: these values are too low IMO. I suggest to change to 1d and 3d

kirg:
About admin vs cli . I think Kobe is working on something to that effect. So not dealing with it in this diff.

kobeyang:
Yes, I'm working on refactoring the cherami-cli and cherami-admin code. Will send out a PR later.

cmd/tools/admin/main.go
datoug: still has dca1 here

PR #247, opened 2017-07-07
Refactor cherami-admin and cherami-cli code

kobeyang:
Basic idea:
1. Create folder common under cmd/tools, and create lib.go in it to store the common flags and commands used by both cli and admin.
2. Delete tools/cli/lib.go and move the corresponding functions to tools/common/lib.go. Add a parameter serviceName to these common functions to differentiate cli and admin.
3. Cherami-cli will just use the common functions and Cherami-admin will use some admin-specific functions in tools/admin/lib.go.

Not sure whether this is the best way to do this, please leave any suggestion whichever you want.


kirg:
@kobeyang i just committed PR #246 that will probably cause some merge conflicts for you .. (sorry about that .. :-))

kobeyang:
Merge with Kiran's change in PR#246.

PR #251, opened 2017-07-17
Replicator: Do not accept SetAckOffset before the cg extent is created locally

datoug:


services/replicator/replicator.go
```diff
@@ -67,6 +67,9 @@ type (
 		storehostConn             map[string]*outConnection
 		storehostConnMutex        sync.RWMutex
 
+		knownCgExtents      map[string]struct{}
```
kirg: May I suggest using the "common/set" library .. it's really just syntactic sugar, but I think it makes code look better.

services/replicator/replicator.go
```diff
@@ -1102,6 +1106,36 @@ func (r *Replicator) SetAckOffset(ctx thrift.Context, request *shared.SetAckOffs
...
+		extent, err := r.metaClient.ReadConsumerGroupExtent(nil, &metadata.ReadConsumerGroupExtentRequest{
+			ConsumerGroupUUID: common.StringPtr(request.GetConsumerGroupUUID()),
+			ExtentUUID:        common.StringPtr(request.GetExtentUUID()),
+		})
+		if err != nil {
+			lcllg.WithField(common.TagErr, err).Error(`SetAckOffset: Failed to read cg extent locally`)
+			r.m3Client.IncCounter(metrics.ReplicatorSetAckOffsetScope, metrics.ReplicatorFailures)
+			return err
+		}
+		if len(extent.GetExtent().GetStoreUUIDs()) < 1 {
```
kirg: "ReadConsumerGroupExtent" (in the metadata client), already checks for this condition .. and returns an error. So the check here is redundant .. and will likely never be reached.

PR #262, opened 2017-07-28
Make all consistency levels for cassandra queries configurable

kirg:


clients/metadata/metadata_cassandra.go
```diff
@@ -189,26 +189,40 @@ type CassandraMetadataService struct {
...
+	// use a minimum default consistency of 'One'
+	lowCons, midCons, highCons = gocql.One, gocql.One, gocql.One
+
+	switch cons := strings.Split(cfgCons, ","); len(cons) {
+	case 1:
+		highCons = gocql.ParseConsistency(strings.TrimSpace(cons[0]))
+
+	case 2:
+		lowCons = gocql.ParseConsistency(strings.TrimSpace(cons[0]))
+		midCons = lowCons
```
boy-uber: It is confusing: when two values [0] is low and [1] is high, when three values, values [0] is low and [1] is mid. Is this to keep backward compatibility?

PR #265, opened 2017-08-02
fix replicator FD leak: client side of websocket needs to drain all messages before returning otherwise the underlying connection will not be closed

datoug:


services/replicator/outconn.go
```diff
@@ -153,94 +153,89 @@ func (conn *outConnection) readMsgStream() {
 	var numMsgsRead int32
 
 	for {
-		select {
-		case <-conn.closeChannel:
+		if numMsgsRead >= creditBatchSize {
```
kirg: I think it will be "cleaner" to move this block of code .. to after the switch block (below) that updates the 'numMsgsRead' counter.

PR #268, opened 2017-08-08
Add logs to debug PurgeMessages latency spikes

kirg:


storage/manyrocks/manyrocks.go
```diff
@@ -827,14 +829,67 @@ func (t *Rock) Purge(purgeAddr s.Address) (nextAddr s.Address, nextKey s.Key, er
...
+					di.deleted = true
+
+				} else {
+					di.deleted = false
+				}
+
+				dbg = append(dbg, di)
+			}
+
+			if duration := time.Since(t0); duration > 2*time.Second {
```
boy-uber: Is it possible that we have too much log here? Maybe adding some limit?

kirg: i think the scale is not very large .. we only log for extents that take a long time and retention runs only once in 10 mins ..

PR #272, opened 2017-08-10
Update cherami-client to v2.3.0

kobeyang:


kirg:
you might want to add a version tag to lock cherami-thrift to v1.24.0-rc0 in your glide.yaml.

glide.yaml
```diff
@@ -32,14 +33,15 @@ import:
...
-  version: =1.23.0
+  version: =2.3.0
   subpackages:
   - client/cherami
   - common
   - common/backoff
   - common/websocket
   - stream
 - package: github.com/uber/cherami-thrift
+  version: v1.24.0-rc0
```
datoug: I would remove the -rc0, to be consistent with the other repos. We can make this a pre-release anyway.

kobeyang: Yes, I prefer we do that next time. Otherwise, we need to tag everything again...

PR #274, opened 2017-08-14
Option to disable NACK throttling for consumer group

kobeyang:


glide.lock
```diff
@@ -134,7 +134,7 @@ imports:
   - common/websocket
   - stream
 - name: github.com/uber/cherami-thrift
-  version: 2cb0e2eeb6570800a2dd86544909bf2693f50e7b
+  version: 3fe5d05728858405b75cd8fc6e1ed7388c5de86e
```
kobeyang: Will update this once the change of cherami-thrift merged.

clients/metadata/metadata_cassandra_test.go
```diff
@@ -52,7 +52,10 @@ type CassandraSuite struct {
 	TestCluster
 }
 
-const testPageSize = 2
+const (
+	testPageSize              = 2
+	FlagDisableNackThrottling = "disable_nack_throttling"
```
thuningxu: This should be a constant somewhere in common.

services/outputhost/messagecache.go
```diff
@@ -973,6 +974,14 @@ func (msgCache *cgMsgCache) updateConn(connID int, event msgEvent, badConns map[
...
 		}
 	} else {
+		if event == eventNACK {
+			// if NACK throttling is disabled for this cg, then no-op
+			throttlingDisabled, ok := msgCache.cgCache.cachedCGDesc.Options[toolscommon.FlagDisableNackThrottling]
+			if ok && throttlingDisabled { // no-op
+				return
+			}
+		}
+
```
thuningxu: Why do it here instead of in `writeMsgsStream()` in `consconnection.go`?

PR #276, opened 2017-08-16
cherami-cli: update seal-check to work on specific destination and DLQs

kirg:


tools/common/lib.go
```diff
@@ -1837,12 +1823,116 @@ iterate_listdestinations_pages:
...
+		req := &shared.ListConsumerGroupRequest{
+			Limit: common.Int64Ptr(DefaultPageSize),
+		}
+
+		for {
+			resp, err1 := mClient.ListAllConsumerGroups(req)
+			ExitIfError(err1)
+
+			for _, cg := range resp.GetConsumerGroups() {
+				if destUUID == cg.GetDestinationUUID() {
```
datoug: Nit: why not filter the destination out in ListAllConsumerGroups call?

datoug: I mean, use ListConsumerGroups

PR #280, opened 2017-08-21
Refactor the replication connections

datoug:
**Problem to solve**
1. FD leak caused by client side not draining from the read stream. (related fixes: https://github.com/uber/cherami-server/pull/265, https://github.com/uber/cherami-server/pull/266)
2. Bug on shutdown path that is causing sealed message not made to store.

**Design philosophies**
read pump reads from stream, write pump writes to stream
read pump communicate with write pump using an internal channel. Read pump writes to the internal channel, and write pump reads from it
Graceful shutdown sequence:
 1. read pump gets a read error (remote shuts down the connection)
 2. read pump closes the internal channel, then exits
 3. write pump detects the internal channel is closed
 4. write pump calls stream.Done(), and exits

**Other changes**
Removed the error handling cases in OutConn.go since same error handling is also done in store. Now replicator serves as a proxy only.

### DIFF
--- a/services/replicator/inconn.go
+++ b/services/replicator/inconn.go
@@ -150,7 +150,11 @@ func (conn *inConnection) writeMsgsStream() {
 			}
 		} else {
 			select {
-			case msg := <-conn.msgCh:
+			case msg, ok := <-conn.msgCh:
+				if !ok {
+					conn.logger.Info("msg channel closed")
+					return
+				}
 				if err := conn.stream.Write(msg); err != nil {
 					conn.logger.Error("write msg failed")
 					return
--- a/services/replicator/inconn.go
+++ b/services/replicator/inconn.go
@@ -94,11 +94,12 @@ func (conn *inConnection) WaitUntilDone() {
 
 func (conn *inConnection) shutdown() {
 	close(conn.shutdownCh)
+	conn.logger.Info(`in connection shutdown`)
 }
 
 func (conn *inConnection) readCreditsStream() {
-	defer close(conn.creditsCh)
 	defer conn.wg.Done()
+	defer close(conn.creditsCh)
 	for {
 		msg, err := conn.stream.Read()
 		if err != nil {
@@ -124,8 +125,8 @@ func (conn *inConnection) readCreditsStream() {
 }
 
 func (conn *inConnection) writeMsgsStream() {
-	defer conn.stream.Done()
 	defer conn.wg.Done()
+	defer conn.stream.Done()
 
 	flushTicker := time.NewTicker(flushTimeout)
 	defer flushTicker.Stop()
@@ -147,6 +148,8 @@ func (conn *inConnection) writeMsgsStream() {
 					conn.logger.Warn("credit flow expired")
 					return
 				}
+			case <-conn.shutdownCh:
+				return
 			}
 		} else {
 			select {
@@ -187,6 +190,8 @@ func (conn *inConnection) writeMsgsStream() {
 					conn.logger.Error(`flush msg failed`)
 					return
 				}
+			case <-conn.shutdownCh:
+				return
 			}
 		}
 	}
--- a/services/replicator/inconn.go
+++ b/services/replicator/inconn.go
@@ -47,7 +47,8 @@ type (
 		creditsCh            chan int32 // channel to pass credits from readCreditsStream to writeMsgsStream
 		creditFlowExpiration time.Time  // credit expiration is used to close the stream if we don't receive any credit for some period of time
 
-		wg sync.WaitGroup
+		wg         sync.WaitGroup
+		shutdownCh chan struct{}
 	}
 )
 
@@ -74,6 +75,7 @@ func newInConnection(extUUID string, destPath string, stream storeStream.BStoreO
 		perDestMetricsScope:  perDestMetricsScope,
 		creditsCh:            make(chan int32, 5),
 		creditFlowExpiration: time.Now().Add(creditFlowTimeout),
+		shutdownCh:           make(chan struct{}),
 	}
 
 	return conn
@@ -88,6 +90,10 @@ func (conn *inConnection) open() {
 
 func (conn *inConnection) WaitUntilDone() {
 	conn.wg.Wait()
+}
+
+func (conn *inConnection) shutdown() {
+	close(conn.shutdownCh)
 }
 
 func (conn *inConnection) readCreditsStream() {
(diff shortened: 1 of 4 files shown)


datoug:
@kirg For the empty credit case(https://github.com/uber/cherami-server/pull/280/files#diff-f8daca02f7ff8d4e7df757fe00f19bc4L148), I think we can rely on the idle timeout(creditFlowTimeout) to eventually close the pumps, instead of introducing another notification channel.

services/replicator/inconn.go
```diff
@@ -75,69 +73,59 @@ func newInConnection(extUUID string, destPath string, stream storeStream.BStoreO
...
+	conn.wg.Wait()
+}
 
-	if !conn.closed {
-		close(conn.closeChannel)
-		conn.closed = true
-	}
-	conn.logger.Info("in connection closed")
+func (conn *inConnection) shutdown() {
+	close(conn.shutdownCh)
```
kirg: You might want to put a log line here .. so you know the reasons that the go-routines have gone, etc.

services/replicator/inconn.go
```diff
@@ -75,69 +73,59 @@ func newInConnection(extUUID string, destPath string, stream storeStream.BStoreO
...
+				WithField(`channelLen`, len(conn.creditsCh)).
+				WithField(`credits`, msg.GetCredits()).
+				Warn(`Dropped credits because of blocked channel`)
 		}
 	}
 }
 
 func (conn *inConnection) writeMsgsStream() {
 	defer conn.stream.Done()
+	defer conn.wg.Done()
```
kirg: you should probably have "defer conn.wg.Done()" _before_ "conn.stream.Done()" ..

services/replicator/inconn.go
```diff
@@ -146,24 +134,29 @@ func (conn *inConnection) writeMsgsStream() {
...
+					return
+				}
 				conn.extentCreditExpiration()
 				localCredits += credit
 			case <-time.After(creditFlowTimeout):
 				conn.logger.Warn("credit flow timeout")
 				if conn.isCreditFlowExpired() {
 					conn.logger.Warn("credit flow expired")
-					go conn.close()
+					return
```
kirg: wait (and quit) on the "shutdownCh" here ..

services/replicator/inconn.go
```diff
@@ -182,16 +175,18 @@ func (conn *inConnection) writeMsgsStream() {
...
+					conn.logger.Info(`internal credit channel closed`)
+					return
+				}
 				conn.extentCreditExpiration()
 				localCredits += credit
 			case <-flushTicker.C:
 				if err := conn.stream.Flush(); err != nil {
 					conn.logger.Error(`flush msg failed`)
-					go conn.close()
+					return
```
kirg: also wait (and return) on the shutdownCh here ..

services/replicator/outconn.go
```diff
@@ -82,43 +99,33 @@ func newOutConnection(extUUID string, destPath string, stream storeStream.BStore
...
+}
 
-	conn.logger.Info("out connection closed")
+func (conn *outConnection) shutdown() {
+	close(conn.shutdownCh)
 }
 
 func (conn *outConnection) writeCreditsStream() {
 	defer conn.stream.Done()
+	defer conn.wg.Done()
```
kirg: as with inconn, move this up one line ..

services/replicator/outconn.go
```diff
@@ -128,116 +135,63 @@ func (conn *outConnection) writeCreditsStream() {
...
-			continue readloop
-
-		case store.ReadMessageContentType_ERROR:
-			msgErr := rmc.GetError()
-			conn.logger.WithField(`Message`, msgErr.GetMessage()).Error(`received error from reading msg`)
-			go conn.close()
-			continue readloop
+		conn.m3Client.IncCounter(conn.metricsScope, metrics.ReplicatorOutConnMsgRead)
+		if rmc.GetType() == store.ReadMessageContentType_SEALED {
+			conn.logger.WithField(`SequenceNumber`, rmc.GetSealed().GetSequenceNumber()).Info(`extent sealed`)
```
kirg: for debug/assertion purposes .. do you want to remember that you saw a "sealed" message .. and in case you don't an error from the next read, log an error or something?

services/replicator/outconn.go
```diff
@@ -128,116 +135,63 @@ func (conn *outConnection) writeCreditsStream() {
...
-			}
-
-			continue readloop
-
-		case store.ReadMessageContentType_ERROR:
-			msgErr := rmc.GetError()
-			conn.logger.WithField(`Message`, msgErr.GetMessage()).Error(`received error from reading msg`)
-			go conn.close()
-			continue readloop
+		conn.m3Client.IncCounter(conn.metricsScope, metrics.ReplicatorOutConnMsgRead)
```
kirg: this will count "SEALED" also as a message .. i think you should not; that way tallying message counts across services will be be easier.

services/replicator/replicator.go
```diff
@@ -258,10 +257,8 @@ func (r *Replicator) OpenReplicationReadStreamHandler(w http.ResponseWriter, req
 	inConn := newInConnection(extUUID, destDesc.GetPath(), inStream, outConn.msgsCh, r.logger, r.m3Client, metrics.OpenReplicationReadScope, metrics.OpenReplicationReadPerDestScope)
 	inConn.open()
 
-	go r.manageInOutConn(inConn, outConn)
-	<-inConn.closeChannel
-	<-outConn.closeChannel
-
+	outConn.WaitUntilDone()
+	inConn.WaitUntilDone()
```
kirg: i might suggest setting up a waitgroup within this function .. and passing a pointer to the waitgroup to inconn/outconn, etc that they increment/decrement .. and you can just do a wg.Wait() here that will automatically wait for both, etc.

services/replicator/replicator.go
```diff
@@ -339,10 +336,8 @@ func (r *Replicator) OpenReplicationRemoteReadStreamHandler(w http.ResponseWrite
 	inConn := newInConnection(extUUID, destDesc.GetPath(), inStream, outConn.msgsCh, r.logger, r.m3Client, metrics.OpenReplicationRemoteReadScope, metrics.OpenReplicationRemoteReadPerDestScope)
 	inConn.open()
 
-	go r.manageInOutConn(inConn, outConn)
-	<-inConn.closeChannel
-	<-outConn.closeChannel
-
+	outConn.WaitUntilDone()
+	inConn.WaitUntilDone()
```
kirg: same as previous comment ..

PR #281, opened 2017-08-25
Hack -> Config: add options to disable smart retry

kobeyang:


tools/common/lib.go
```diff
@@ -503,11 +503,17 @@ func CreateConsumerGroupSecure(
 	isMultiZone := len(zoneConfigs.GetConfigs()) > 0
 
 	options := make(map[string]string)
-	disableNackThrottling := strings.ToLower(c.String(common.FlagDisableNackThrottling))
-	if disableNackThrottling == "true" {
+
+	disableNackThrottling := c.Bool(common.FlagDisableNackThrottling)
+	if disableNackThrottling == true {
```
kirg: you could probably simply do:
```
if c.Bool(common.FlagDisableNackThrottling {
   ...
```
instead of assigning it to a separate variable just for checking, etc.

tools/common/lib.go
```diff
@@ -503,11 +503,17 @@ func CreateConsumerGroupSecure(
...
 		options[common.FlagDisableNackThrottling] = "true"
 	}
 
+	disableSmartRetry := c.Bool(common.FlagDisableSmartRetry)
+	if disableSmartRetry == true {
```
kirg: same as above.

Issue #284, opened 2017-08-30
[QUESTION] How to get Cherami to Spread the Load


djoodle:
I'm running  a Cherami cluster using docker containers. I have:

- 3x Frontend Hosts (0.5GB mem)
- 3x Controllers (0.5GB mem)
- 3x Inputhosts (2GB mem)
- 3x Outputhosts (2GB mem)
- 5x Storage hosts (4GB mem)

And Cassandra with 5 nodes, and a RF of 3.

However when we have times of heavy load across the system, most of the boxes sit idle with one storage host that climbs to all its memory and then is killed (OOM). 

Is there a way to tell Cherami to spread the load to the other storage hosts? It seems all the extents are placed on the one host, and even when that hosts starts slowing down and getting behind, the load is never shared.

Cassandra is aware of all nodes, each node seems to be working

Many thanks!

danudell-trustnetworks:
For reference my current configuration is:

```
DefaultServiceConfig:
  ListenAddress: "XXXX"
  RingHosts: "frontend-0.cherami:4922,frontend-1.cherami:4922,frontend-2.cherami:4922"
  EnableLimits: true

# ServiceConfig overrides default config with service-specific config such as ports
ServiceConfig:
  cherami-inputhost:
    Port: 4240
    WebsocketPort: 6189
    tchannel:
      port: 4240
      disableHyperbahn: true
      logLevel: warn
      disableLogging: true
  cherami-storehost:
    Port: 4253
    WebsocketPort: 6191
    tchannel:
      port: 4253
      disableHyperbahn: true
      logLevel: warn
      disableLogging: true
  cherami-outputhost:
    Port: 4254
    WebsocketPort: 6190
    tchannel:
      port: 4254
      disableHyperbahn: true
      logLevel: warn
      disableLogging: true
  cherami-frontendhost:
    Port: 4922
    tchannel:
      port: 4922
      disableHyperbahn: true
      logLevel: warn
      disableLogging: false
  cherami-controllerhost:
    StorePlacementConfig:
      MinFreeDiskSpaceBytes: 10000
    Port: 5425
    tchannel:
      port: 5425
      disableHyperbahn: true
      logLevel: warn
      disableLogging: false
  cherami-replicator:
    Port: 6280
    WebsocketPort: 6310
    tchannel:
      port: 6280
      disableHyperbahn: true
      logLevel: warn
      disableLogging: false

DefaultDestinationConfig:
  Replicas: 1

MetadataConfig:
[...]

PR #285, opened 2017-08-30
pick host randomly when distance map is empty

datoug:
Currently when distance map is empty, we always pick the first N hosts. This patch fixes the issue by randomly picking hosts. I suspect this is issue is the cause of https://github.com/uber/cherami-server/issues/284

### DIFF
--- a/services/controllerhost/placement.go
+++ b/services/controllerhost/placement.go
@@ -34,6 +34,7 @@ import (
 )
 
 var errNoHosts = errors.New("Unable to find healthy hosts")
+var errNotEnoughHosts = errors.New("Unable to enough hosts")
 var errNoInputHosts = errors.New("Unable to find healthy input host")
 var errNoOutputHosts = errors.New("Unable to find healthy output host")
 var errNoStoreHosts = errors.New("Unable to find healthy store hosts")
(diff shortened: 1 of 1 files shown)


services/controllerhost/placement.go
```diff
@@ -77,7 +78,17 @@ func toResources(hosts []*common.HostInfo) []string {
...
-		return poolHosts[:count], nil
+		if cnt := len(poolHosts); cnt >= count {
+			start := rand.Intn(cnt)
+			end := start + count
+			if end > cnt {
+				end = cnt
+			}
+			pickedHosts := append([]*common.HostInfo{}, poolHosts[start:end]...)
+			pickedHosts = append(pickedHosts, poolHosts[:count-len(pickedHosts)]...)
+			return pickedHosts, nil
```
kirg: The above picks hosts from a random starting point, in a fixed order. It would probably be better to pick a truly random set of hosts:
```
pickedHosts := make([]*common.HostInfo, count)
for i, j := rand.Perm(len(poolHosts)) {
    if i >= count {
        break
    }
    pickedHosts[i] = poolHosts[j]
}
return pickedHosts, nil
```

datoug: this is awesome..

danudell-trustnetworks:
Are you able to reinstate the branch? Since its been reverted back? Alternatively have you can an example of a distance mapping where all hosts are set to equidistant from each other that we could use? Would that work?

PR #290, opened 2017-09-11
Slowdown redeliveries, add timeout metrics

kobeyang:
Talked with Kiran, set the nack redelivery delay to 25 milliseconds for now. We can tune it later.

### DIFF
--- a/services/outputhost/messagecache.go
+++ b/services/outputhost/messagecache.go
@@ -165,6 +165,7 @@ const defaultLockTimeoutInSeconds int32 = 42
 const defaultMaxDeliveryCount int32 = 2
 const blockCheckingTimeout time.Duration = time.Minute
 const redeliveryInterval = time.Second / 2
+const nackRedeliveryDelayInMillisecond int32 = 25
 
 // SmartRetryDisableString can be added to a destination or CG owner email to request smart retry to be disabled
 // Note that Google allows something like this: gbailey+smartRetryDisable@uber.com
@@ -329,8 +330,8 @@ func (msgCache *cgMsgCache) utilHandleDeliveredMsg(cMsg cacheMsg) {
 		msgCache.changeState(ackID, stateConsumed, msg, eventCache)
 	case stateEarlyNACK:
 		//lclLg.WithField("AckID", common.ShortenGUIDString(msg.GetAckId())).Debug("manageMessageDeliveryCache: Early NACKed message (delivering to DLQ)")
-		msgCache.changeState(ackID, stateDelivered, msg, eventCache) // Mark one delivery as complete, eligible for redelivery depending on max deliveries
-		cm.fireTime = msgCache.addTimer(0, ackID)                    // Try to redeliver immediately, rather than wait for the lock timeout
+		msgCache.changeState(ackID, stateDelivered, msg, eventCache)             // Mark one delivery as complete, eligible for redelivery depending on max deliveries
+		cm.fireTime = msgCache.addTimer(nackRedeliveryDelayInMillisecond, ackID) // Try to redeliver immediately, rather than wait for the lock timeout
 	case stateDelivered:
 		break // this happens on redelivery
 	case stateConsumed:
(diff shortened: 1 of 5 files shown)


services/outputhost/messagecache.go
```diff
@@ -537,18 +548,18 @@ func (msgCache *cgMsgCache) utilRenewCredits() {
 	}
 }
 
-func (msgCache *cgMsgCache) addTimer(delaySeconds int, id AckID) common.UnixNanoTime {
+func (msgCache *cgMsgCache) addTimer(delayMilliSeconds int32, id AckID) common.UnixNanoTime {
 	now := common.Now()
 	entry := &timerCacheEntry{
 		AckID:    id,
-		fireTime: now + common.UnixNanoTime(int64(time.Second)*int64(delaySeconds)),
+		fireTime: now + common.UnixNanoTime(int64(time.Second)*int64(delayMilliSeconds)),
```
kirg: multiply by time.Millisecond (instead of time.Second).

services/outputhost/messagecache.go
```diff
@@ -537,18 +548,18 @@ func (msgCache *cgMsgCache) utilRenewCredits() {
 	}
 }
 
-func (msgCache *cgMsgCache) addTimer(delaySeconds int, id AckID) common.UnixNanoTime {
+func (msgCache *cgMsgCache) addTimer(delayMilliSeconds int32, id AckID) common.UnixNanoTime {
```
kirg: probably better to take delay of type "time.Duration" instead of "int32" milliseconds, etc. and at the call-site you could just multiply by "time.Second", etc; and re-define "nackRedeliveryDelayInMillisecond" as simply "nackRedeliveryDelay" (that can be expressed in time.Duration).

services/outputhost/messagecache.go
```diff
@@ -401,7 +403,7 @@ func (msgCache *cgMsgCache) utilHandleRedeliveryTicker() {
 		msgCache.reinjectlastAckMsg() // injects the last acked message if it is not already injected
 	}
 
-	for _, cache := range timerCaches {
+	for i, cache := range timerCaches {
```
kirg: if i am reading it right .. it seems like this function (utilHandleRedeliveryTicker) gets called once in 500ms (redeliveryInterval). for our nack-redelivery case, it seems like if the message is not ready for redelivery .. it will be checked again only after 500ms, even though our nack-redelivery delay is just 25ms.

services/outputhost/messagecache.go
```diff
@@ -979,12 +990,20 @@ func (msgCache *cgMsgCache) updateConn(connID int, event msgEvent) {
 
 func (msgCache *cgMsgCache) refreshCgConfig(oldOutstandingMessages int32) {
 	outstandingMsgs := oldOutstandingMessages
+	redeliveryIntervalInMs := msgCache.redeliveryIntervalInMs
+
 	cfg, err := msgCache.cgCache.getDynamicCgConfig()
 	if err == nil {
 		outstandingMsgs = msgCache.cgCache.getMessageCacheSize(cfg, oldOutstandingMessages)
+		redeliveryIntervalInMs = cfg.RedeliveryIntervalInMs
```
kirg: ideally, this should made be a "per CG" config, like the "message cache size" config.

services/outputhost/messagecache.go
```diff
@@ -979,12 +990,20 @@ func (msgCache *cgMsgCache) updateConn(connID int, event msgEvent) {
...
 	}
 
 	msgCache.maxOutstandingMsgs = outstandingMsgs
+
+	if redeliveryIntervalInMs != msgCache.redeliveryIntervalInMs {
+		msgCache.redeliveryTicker = time.NewTicker(time.Duration(redeliveryIntervalInMs) * time.Millisecond)
```
kirg: before creating a new ticker, you should "stop the ticker to release associated resources" (https://golang.org/pkg/time/#NewTicker).

kobeyang: Nice!

PR #295, opened 2017-09-20
Factor out kafka stream converter interface

mmchen:
Factor out kafka stream converter to make it possible for injecting customized converter


services/outputhost/kafkaStreamConverter.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/uber-common/bark"
+	"github.com/uber/cherami-server/common"
+	"github.com/uber/cherami-thrift/.generated/go/cherami"
+	"github.com/uber/cherami-thrift/.generated/go/store"
+)
+
+// KafkaStreamConverter defines interface to convert a Kafka message to Cherami message
+type KafkaStreamConverter interface {
+	convertKafkaMessageToCherami(m *s.ConsumerMessage, logger bark.Logger) (c *store.ReadMessageContent)
+}
```
kirg: like i mentioned in the other comment .. perhaps it would be better to provide a "factory" interface that returns a converter function based on various details about the destination/CG/kafka topics, etc.
```
type KafkaMessageTranslatorConfig struct {
    Destination *cherami.DestinationDescription
    ConsumerGroup *cherami.ConsumerGroupDescription
    KafkaTopics []string // optional: already contained in destination-description
    KafkaCluster string  // optional: already contained in destination-description
    // FIXME: any other useful args; say pointer to sarama consumer?
}

type KafkaMessageTranslator func(kMsg *sarama.ConsumerMessage) (cMsg *store.ReadMessageContent)

type KafkaMessageTranslatorFactory interface {
    GetTranslator(cfg *KafkaMessageTranslatorConfig, log bark.Logger) KafkaMessageTranslator
}
```

Btw, another option to consider is if it might help to push this interface one layer up .. and provide a factory that gets us an implementation of the "BStoreOpenAppendStreamOutCall" interface. That might give a lot more flexibility and perhaps even help with, say, transitioning consumers of the 'cherami-ingestor' destinations to KFC? I think you might know more about this than I do .. let's discuss this offline.

services/outputhost/kafkaStream.go
```diff
@@ -33,17 +33,37 @@ import (
...
 	logger          bark.Logger
 	seqNo           int64
 }
 
-var errKafkaNilControlFlow = errors.New(`nil or non-positive controlFlow passed to Write()`)
-var errKafkaClosed = errors.New(`closed`)
+// KafkakMessageConverterConfig is used to config customized converter
+type KafkaMessageConverterConfig struct {
+	Destination *cherami.DestinationDescription
+	ConsumerGroup *cherami.ConsumerGroupDescription
```
kirg: since we are not filling these in currently, probably better to comment them out for now.

PR #296, opened 2017-09-21
Introduce consumer-group "deleting" status

kirg:
This diff handles adding a "deleting" (pre-delete) status to consumer-group. The actual deletion of the consumer-group would be done in Retention Manager for now .. and I will do that in a separate diff.

### DIFF
--- a/clients/metadata/metadata_cassandra.go
+++ b/clients/metadata/metadata_cassandra.go
@@ -1634,7 +1634,8 @@ func (s *CassandraMetadataService) UpdateConsumerGroup(ctx thrift.Context, reque
 		return nil, err
 	}
 
-	if existingCG.GetStatus() == shared.ConsumerGroupStatus_DELETED {
+	if existingCG.GetStatus() == shared.ConsumerGroupStatus_DELETING ||
+		existingCG.GetStatus() == shared.ConsumerGroupStatus_DELETED {
 		return nil, &shared.BadRequestError{
 			Message: fmt.Sprintf("UpdateConsumerGroup - Attempt to update DELETED consumer group, dst=%v, cg=%v",
 				request.GetDestinationPath(), request.GetConsumerGroupName()),
@@ -1748,7 +1749,8 @@ func (s *CassandraMetadataService) DeleteConsumerGroup(ctx thrift.Context, reque
 		}
 	}
 
-	if existingCG.GetStatus() == shared.ConsumerGroupStatus_DELETED {
+	if existingCG.GetStatus() == shared.ConsumerGroupStatus_DELETING ||
+		existingCG.GetStatus() == shared.ConsumerGroupStatus_DELETED {
 		return nil
 	}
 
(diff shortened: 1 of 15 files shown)


glide.yaml
```diff
@@ -40,7 +40,7 @@ import:
   - common/websocket
   - stream
 - package: github.com/uber/cherami-thrift
-  version: v1.25.0-rc0
+  version: 45b0a9da65195b298b3967360a1a069907d5266a
```
datoug: Nit: You can remove this.

clients/metadata/metadata_cassandra.go
```diff
@@ -1840,6 +1843,68 @@ func (s *CassandraMetadataService) DeleteConsumerGroup(ctx thrift.Context, reque
 	return nil
 }
 
+// DeleteConsumerGroupUUID deletes the consumer-group corresponding to the given UUID
+// from the consumer_groups table
+func (s *CassandraMetadataService) DeleteConsumerGroupUUID(ctx thrift.Context, request *m.DeleteConsumerGroupUUIDRequest) error {
```
datoug: so dumb question, why this new API is needed?

services/replicator/metadataReconciler.go
```diff
@@ -303,9 +303,9 @@ func (r *metadataReconciler) reconcileCg(localCgs []*shared.ConsumerGroupDescrip
 		})
 		localCg, ok := localCgsSet[remoteCg.GetConsumerGroupUUID()]
 		if ok {
-			if remoteCg.GetStatus() == shared.ConsumerGroupStatus_DELETED {
+			if remoteCg.GetStatus() == shared.ConsumerGroupStatus_DELETING {
```
datoug: I think we should check for both condition here. If remote is in deleting or deleted, and local is not in either of the state, we should call deleteConsumergroup.

services/replicator/metadataReconciler.go
```diff
@@ -333,7 +333,7 @@ func (r *metadataReconciler) reconcileCg(localCgs []*shared.ConsumerGroupDescrip
 			replicatorReconcileCgFoundMissingCount = replicatorReconcileCgFoundMissingCount + 1
 
 			// If the missing ConsumerGroup is in deleted status, we don't need to create the ConsumerGroup locally
-			if remoteCg.GetStatus() == shared.ConsumerGroupStatus_DELETED {
+			if remoteCg.GetStatus() == shared.ConsumerGroupStatus_DELETING {
```
datoug: I think we should check for both conditions(deleted/deleting). If remote is in deleted, we also don't need to create it locally.

clients/metadata/metadata_cassandra.go
kirg: Since the "DeleteConsumerGroup" API now only moves the CG to deleting state, this new API is needed to move the CG to _deleted_ state (and apply TTL to the row in the consumer_groups table, etc). Basically this would do the actual cleanup of the row .. and would be called by retention manager when the CG is ready to be actually purged.

services/replicator/metadataReconciler.go
kirg: CGs in "DELETED" state are *not* returned by ListConsumerGroups .. so this code was dead code earlier. It however does return CGs in "DELETING" state .. and the code below that calls "DeleteConsumerGroup" locally would then move the local CG to DELETING, as well.

services/replicator/metadataReconciler.go
kirg: CGs in "DELETED" state are *not* returned by the ListConsumerGroups API .. and are technically 'invisible'. So that check (on a DELETED CG) would be pointless.

PR #297, opened 2017-09-25
replicator: clear the client cache when host list changes

datoug:
When a host list changes for a deployment, we should clear the client cache so that the new list can be used for new request.

### DIFF
--- a/services/replicator/replicatorclientfactory.go
+++ b/services/replicator/replicatorclientfactory.go
@@ -171,8 +171,9 @@ func (f *repCltFactoryImpl) GetHostsForAllDeployment() map[string][]string {
 
 func (f *repCltFactoryImpl) UpdateHostsForDeployment(deployment string, hosts []string) {
 	f.hostLk.Lock()
-	f.deploymentClientsLk.Lock()
 	defer f.hostLk.Unlock()
+
+	f.deploymentClientsLk.Lock()
 	defer f.deploymentClientsLk.Unlock()
 
 	f.hosts[deployment] = hosts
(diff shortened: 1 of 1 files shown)


services/replicator/replicatorclientfactory.go
```diff
@@ -170,7 +171,12 @@ func (f *repCltFactoryImpl) GetHostsForAllDeployment() map[string][]string {
 
 func (f *repCltFactoryImpl) UpdateHostsForDeployment(deployment string, hosts []string) {
 	f.hostLk.Lock()
+	f.deploymentClientsLk.Lock()
+	defer f.deploymentClientsLk.Unlock()
 	defer f.hostLk.Unlock()
```
kirg: you probably want to move the 'defer hostLk.Unlock()' to be immediately after the "h.hostLk.Lock()' statement. (the 'defers' are processed LIFO order ..)

services/replicator/replicatorclientfactory.go
```diff
@@ -65,11 +65,12 @@ type (
...
-		logger            bark.Logger
-		ch                *tchannel.Channel
-		deploymentClients map[string][]replicator.TChanReplicator
-		lk                sync.RWMutex
+		AppConfig configure.CommonAppConfig
+		logger    bark.Logger
+		ch        *tchannel.Channel
+
+		deploymentClients   map[string][]replicator.TChanReplicator
+		deploymentClientsLk sync.RWMutex
```
kirg: Good news: Go 1.9 has concurrent maps! https://golang.org/pkg/sync/#Map

Bad news: we can't use it yet .. since we probably want to maintain support for Go 1.8 too.

services/replicator/replicatorclientfactory.go
```diff
@@ -172,8 +172,8 @@ func (f *repCltFactoryImpl) GetHostsForAllDeployment() map[string][]string {
 func (f *repCltFactoryImpl) UpdateHostsForDeployment(deployment string, hosts []string) {
 	f.hostLk.Lock()
 	f.deploymentClientsLk.Lock()
-	defer f.deploymentClientsLk.Unlock()
 	defer f.hostLk.Unlock()
+	defer f.deploymentClientsLk.Unlock()
```
kirg: Ideally, best to have the 'defer lk.Unlock()" immediately follow the line that does "lk.Lock()".

PR #302, opened 2017-09-29
fix lint error in travis-ci

mmchen:


kirg:
@mmchen, i think lint is still throwing an error:
```
./services/outputhost/kafkaStream.go:46:2: don't use leading k in Go names; struct field kConverter should be converter
./services/outputhost/kafkaStream.go:51:1: comment on exported type KafkaMessageConverterConfig should be of the form "KafkaMessageConverterConfig ..." (with optional leading article)
```

PR #305, opened 2017-10-04
[prototype] size-capped channel

kirg:
Prototype implementation of a size-capped channel (szChan). This could be useful in 'outputhost' where we want to move to size-based restriction .. as opposed to how it is done currently, based on message counts. This will enable us to more gracefully handle super-large message sizes, etc.

When you create a size-capped channel, you specify the 'capacity' in size/bytes .. and it returns two go-channels, one for sending (writer) into it the size-capped channel and another for receiving (reader) from it. Each time a message is sent to the szChan it buffers it and keeps a tally of the buffered-size -- if that exceeds the capacity, it would block sends, until space is made .. by reading out messages from it.

Bonus: Unlike count-capped go-channels, you can change the capacity of the size-based channel dynamically!

NB: Since there isn't a means to 'peek' into a go-channel, we would need to read out a message before being able to determine whether it would fit or not. So the channel would block on the message _after_ the one that makes it go beyond capacity. But this should not matter when there's a a continuous stream of messages. Also note, to improve performance, you can also fine-tune the 'buffer'/capacity of the reader/writer channels.

### DIFF
--- a/common/sizedCh/sizedCh.go
+++ b/common/sizedCh/sizedCh.go
@@ -30,6 +30,7 @@ func New(size int) (*SizedCh, <-chan Message, chan<- Message) {
 		readerCh: make(chan unsafe.Pointer, readChSize),
 		writerCh: make(chan unsafe.Pointer, writeChSize),
 		bufferCh: make(chan unsafe.Pointer, bufChSize),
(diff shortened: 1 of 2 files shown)


thuningxu:
I think we decided not to do this.

PR #310, opened 2017-10-10
RetMgr: Handle cleanup of consumer-groups and consumer-group extents

kirg:
This diff also adds a new API to list consumer-groups for a given destination-uuid -- the difference between this and the existing ListConsumerGroups API is that this one would look at the 'consumer_groups' table (instead of 'consumer_groups_by_name', that ListConsumerGroups look at).

Retention-manager uses this new API to query consumer-groups in 'deleting' state and takes appropriate action to 'delete' them.

### DIFF
--- a/clients/metadata/metadata_cassandra.go
+++ b/clients/metadata/metadata_cassandra.go
@@ -1189,6 +1189,22 @@ const (
 		columnZoneConfigs + `: ?, ` +
 		columnOptions + `: ? }`
 
+	cqlConsumerGroupType = columnConsumerGroup + `.` + columnUUID + "," +
+		columnConsumerGroup + `.` + columnDestinationUUID + "," +
+		columnConsumerGroup + `.` + columnName + "," +
+		columnConsumerGroup + `.` + columnStartFrom + "," +
+		columnConsumerGroup + `.` + columnStatus + "," +
+		columnConsumerGroup + `.` + columnLockTimeoutSeconds + "," +
+		columnConsumerGroup + `.` + columnMaxDeliveryCount + "," +
+		columnConsumerGroup + `.` + columnSkipOlderMessagesSeconds + "," +
+		columnConsumerGroup + `.` + columnDelaySeconds + "," +
+		columnConsumerGroup + `.` + columnDeadLetterQueueDestinationUUID + "," +
+		columnConsumerGroup + `.` + columnOwnerEmail + "," +
+		columnConsumerGroup + `.` + columnIsMultiZone + "," +
+		columnConsumerGroup + `.` + columnActiveZone + "," +
+		columnConsumerGroup + `.` + columnZoneConfigs + "," +
+		columnConsumerGroup + `.` + columnOptions
+
 	sqlInsertCGByUUID = `INSERT INTO ` + tableConsumerGroups +
 		`(` +
 		columnUUID + `, ` +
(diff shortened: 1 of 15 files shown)


services/retentionmgr/metadataDep.go
```diff
@@ -60,7 +60,7 @@ func (t *metadataDepImpl) GetDestinations() (destinations []*destinationInfo) {
 	i := 0
 	for {
 
-		log.Debug("GetDestinations: ListDestinationsByUUID on metadata")
+		log.Info("GetDestinations: ListDestinationsByUUID on metadata")
```
datoug: I think these will cause too many logs

services/retentionmgr/metadataDep.go
```diff
@@ -312,19 +312,60 @@ func (t *metadataDepImpl) MarkExtentConsumed(destID destinationID, extID extentI
...
+	})
+
+	for {
+		log.Info("GetExtentsForConsumerGroup: ReadConsumerGroupExtentsLite on metadata")
+
+		res, e := t.metadata.ReadConsumerGroupExtentsLite(ctx, req)
+		if e != nil {
+			err = e
+			log.WithField(common.TagErr, err).Error("GetExtentsForConsumerGroup: ReadConsumerGroupExtentsLite failed")
+			break
```
datoug: return here?

services/retentionmgr/metadataDep.go
```diff
@@ -348,7 +390,33 @@ func (t *metadataDepImpl) DeleteConsumerGroupExtent(cgID consumerGroupID, extID
...
 
-	log.Debug("DeleteConsumerGroupExtent done")
+	log.Info("DeleteConsumerGroupExtent done")
+	return nil
+}
+
+func (t *metadataDepImpl) DeleteConsumerGroup(destID destinationID, cgID consumerGroupID) error {
+
+	req := metadata.NewDeleteConsumerGroupUUIDRequest()
+	req.UUID = common.StringPtr(string(destID))
```
datoug: uuid should be cg uuid right?

clients/metadata/metadata_cassandra.go
```diff
@@ -1189,9 +1189,26 @@ const (
...
+		columnConsumerGroup + `.` + columnZoneConfigs + "," +
+		columnConsumerGroup + `.` + columnOptions
+
 	sqlInsertCGByUUID = `INSERT INTO ` + tableConsumerGroups +
 		`(` +
 		columnUUID + `, ` +
+		columnDestinationUUID + `, ` +
 		columnIsMultiZone + `, ` +
 		columnConsumerGroup +
 		`) VALUES (?, ?,` + sqlCGValue + `)`
```
datoug: You need one more '?' here

datoug: some test should be able to catch this.

services/retentionmgr/retention.go
```diff
@@ -342,10 +435,9 @@ func (t *RetentionManager) runRetention(jobsC chan<- *retentionJob) bool {
 			totalJobs++
 		}
 
-		if allExtentsDeleted && dest.status == shared.DestinationStatus_DELETING {
+		if dest.status == shared.DestinationStatus_DELETING && numCGs == 0 && numExtents == 0 {
```
datoug: I don't think you can rely on 'numCGs == 0' since the error from the below call was swallowed.
cgs := t.metadata.GetConsumerGroups(dest.id)

clients/metadata/metadata_cassandra.go
```diff
@@ -2003,6 +1987,73 @@ func (s *CassandraMetadataService) ListConsumerGroups(ctx thrift.Context, reques
 	return result, nil
 }
 
+// ListConsumerGroupsUUID returns all ConsumerGroups matching the given destination-uuid.
+func (s *CassandraMetadataService) ListConsumerGroupsUUID(ctx thrift.Context, request *shared.ListConsumerGroupsUUIDRequest) (*shared.ListConsumerGroupsUUIDResult_, error) {
```
datoug: can you add some UT for the new API?

services/retentionmgr/retention.go
```diff
@@ -312,26 +314,117 @@ func (t *RetentionManager) runRetention(jobsC chan<- *retentionJob) bool {
 			continue
 		}
 
+		log := t.logger.WithField(common.TagDst, dest.id)
```
datoug: Is it possible to UT the new behavior

clients/metadata/metadata_cassandra.go
```diff
@@ -1189,9 +1189,26 @@ const (
 		columnZoneConfigs + `: ?, ` +
 		columnOptions + `: ? }`
 
+	cqlConsumerGroupType = columnConsumerGroup + `.` + columnUUID + "," +
```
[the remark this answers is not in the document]
kirg: true .. :-)

clients/metadata/metadata_cassandra.go
kirg: i have now.

services/retentionmgr/metadataDep.go
kirg: i do a 'break' so it will print the "done" log below ..

services/retentionmgr/metadataDep.go
kirg: yes; fixed.

services/retentionmgr/metadataDep.go
kirg: not sure now .. why i decided to enable these. reverting it. :-)

services/retentionmgr/retention.go
kirg: it can be .. though it will be quite some work; will give it a shot.

clients/metadata/metadata_cassandra_test.go
```diff
@@ -2056,48 +2106,80 @@ func (s *CassandraSuite) TestListConsumerGroups() {
...
-	inputs := make([]*shared.ListConsumerGroupRequest, 0, 2)
-	inputs = append(inputs, &shared.ListConsumerGroupRequest{
+	assert.True(listCGs(&shared.ListConsumerGroupRequest{
 		DestinationPath: common.StringPtr(dstPath),
 		Limit:           common.Int64Ptr(testPageSize),
-	})
-	inputs = append(inputs, &shared.ListConsumerGroupRequest{
+	}).Equals(cgSet), "ListConsumerGroups did not return all CGs")
+
+	assert.True(listCGs(&shared.ListConsumerGroupRequest{
```
datoug: Is this section duplicated from previous one?

kirg: yes. removed.

services/retentionmgr/retention.go
datoug: I still don't see this get fixed.

services/retentionmgr/retention.go
kirg: as discussed offline, i am going to take this up separately.

PR #315, opened 2017-10-18
Specify consistency level for all the operations that are not specifi…

kobeyang:
…ed before


clients/metadata/metadata_cassandra.go
```diff
@@ -1817,6 +1827,8 @@ func (s *CassandraMetadataService) DeleteConsumerGroup(ctx thrift.Context, reque
 		existingCG.GetDestinationUUID(),
 		existingCG.GetConsumerGroupName())
 
+	batch.Cons = s.lowConsLevel
```
kirg: use 'midConsLevel'

PR #317, opened 2017-10-30
Log metadata queries with high latency

kirg:


common/metadata/metaMetrics.go
```diff
@@ -29,6 +32,11 @@ import (
 	"github.com/uber/tchannel-go/thrift"
 )
 
+const (
+	pointQueryHighLatencyThreshold = 3 * time.Second
+	listQueryHighLatencyThreshold  = 8 * time.Second
```
thuningxu: Seems too high. How about 1 second for point queries and 5 seconds for list queries?

kirg: as discussed offline .. we'll go with these values for now; they are 'high' thresholds, but this is to prevent swamped with logs when there is an ongoing metadata issue. we will experiment to see whether we need to further refine this with one of the following:1
1. rate-limit the logs, so we don't log too much
2. make the timeouts dynamically configurable
3. emit these with destination/cg tags, and perhaps avoid logs all together

PR #321, opened 2017-11-14
storehost: more debug logs

kirg:


kirg:
Discarding; changes subsumed in #323.

PR #322, opened 2017-11-14
[ON-HOLD] Fixes around read-only storehost (on low disk space)

kirg:
- notify controller when storehost is in read-only
- controller would stop placing new extents on stores that are read-only
- make threshold free-space based (instead of available space percentage)
- improve logging on OpenAppendStream failures
- (minor annoyance) rename 'storagemonitor" to "spaceMon" to reduce file completion collision on 'stor<tab>' with "storehost.go". :-)


thuningxu:
Can we separate refactoring from the logic change? This logic is rather important to get right and reviewed throughly. Suggestion:

PR 1:
* make threshold free-space based (instead of available space percentage)

PR 2: 
* notify controller when storehost is in read-only
* controller would stop placing new extents on stores that are read-only

PR 3:
rest

kirg:
okay .. will try and split up the PR.

kirg:
I have created two new PRs:
#325: prevents controller from placing new extents on read-only stores
#326: makes storehost read-only protection based on absolute free-space, instead of percentages

After the above have been handled, I will tackle these other changes that were part of this PR:
- improve logging on OpenAppendStream failures
- (minor annoyance) rename 'storagemonitor" to "spaceMon" to reduce file completion collision on 'stor' with "storehost.go". :-)

kirg:
closing out this PR, since there's too many conflicts with master, etc. will recreate PR with the rest of the changes ..

PR #323, opened 2017-11-15
Plug hole that allowed concurrent replication streams

kirg:
- fix for concurrent replication streams
- update the 'first seqnum' in replication code-path (this was missing)
- added more logs around open/close extent

### DIFF
--- a/services/storehost/extMgr.go
+++ b/services/storehost/extMgr.go
@@ -378,8 +378,7 @@ func (xMgr *ExtentManager) closeExtent(ext *extentContext, intent OpenIntent, op
 		xMgr.hostMetrics.Decrement(load.HostMetricNumReadConns)
 
 	case OpenIntentReplicateExtent:
-
-		if openError != nil {
+		if openError == nil {
 			// mark as not open for replication
 			atomic.StoreInt32(&ext.openedForReplication, 0)
 		}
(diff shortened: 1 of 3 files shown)


services/storehost/extMgr.go
```diff
@@ -363,12 +372,17 @@ func (xMgr *ExtentManager) closeExtent(ext *extentContext, intent OpenIntent, op
...
 		ext.extMetrics.Decrement(load.ExtentMetricNumReadConns)
 		xMgr.hostMetrics.Decrement(load.HostMetricNumReadConns)
+
 	case OpenIntentReplicateExtent:
-		// mark as not open for replication
-		atomic.StoreInt32(&ext.openedForReplication, 0)
+
+		if openError != nil {
+			// mark as not open for replication
+			atomic.StoreInt32(&ext.openedForReplication, 0)
```
kirg: ^^ the fix

thuningxu: if openError == nil?

thuningxu:
Can you add unit test?

kirg:
There is a unit-test that checks to see that when two OpenReplicateStream requests come in, only one will succeed and the other will fail. Unfortunately, this bug shows up when a third OpenReplicateStream request comes in!

kirg:
@thuningxu I have updated the existing test to try another (third) replicate-extent request, which should fail; verified that it fails without fix and passes with.

PR #325, opened 2017-11-17
Stop controller from placing extents on readonly stores

kirg:
- storehosts that are 'read-only' (due to low-space), would emit a (new) node-metric to controller
- controller would stop placing extents on stores that are read-only

This change relies on a thrift change, tracked by this PR:
https://github.com/uber/cherami-thrift/pull/29


kirg:
@thuningxu No, I couldn't find any unit tests that tests placement specifically.

PR #326, opened 2017-11-17
Storehost: make readonly if free-space below threshold

kirg:


services/storehost/storagemonitor.go
```diff
@@ -71,35 +87,39 @@ type (
...
-	go s.doHouseKeeping()
+	go func() {
+
+		ticker := time.NewTicker(storageMonitoringInterval)
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-ticker.C:
+				go s.checkStorage()
```
thuningxu: Why spin another goroutine?

kirg: Well, that's how it used to be ..

In my slightly 'cleaned up' version in PR #322 (https://github.com/uber/cherami-server/pull/322/files#diff-cb966d6a28a81db47b41fccd069cfe0b), I don't spin up a new go-routine.

PR #327, opened 2017-11-19
Sequence number should be outside of message converter

thuningxu:


services/outputhost/kafkaStream.go
```diff
@@ -145,24 +151,23 @@ func GetDefaultKafkaMessageConverter(seqNo *int64, logger bark.Logger) KafkaMess
...
+				Data: m.Value,
+				UserContext: map[string]string{
+					`key`:       string(m.Key),
+					`topic`:     m.Topic,
+					`partition`: strconv.Itoa(int(m.Partition)),
+					`offset`:    strconv.Itoa(int(m.Offset)),
+				},
+				// TODO: Checksum?
+			},
 		}
```
kirg: (minor) why don't we set all of these members (c.Message, c.Message.Message, etc) inline .. as part of initialization in line 131 itself.

kirg:
Good catch. Minor comment, otherwise looks good to me.

thuningxu:
As the title

PR #328, opened 2017-11-20
Clean up logging, 'storagemonitor', etc

kirg:
* improve logging on OpenAppendStream, OpenReadStream failures
* (minor annoyance) rename 'storagemonitor" to "spaceMon" to reduce file completion collision on 'stor' with "storehost.go". :-)


kirg:
discarding .. split into #329 and #330.

PR #330, opened 2017-11-20
rename/clean-up storageMonitor -> spaceMon

kirg:


services/storehost/spaceMon.go
```diff
@@ -31,20 +31,30 @@ import (
...
+var (
+	// TODO: make these dynamically configurable!
+
+	// SpaceMonThresholdWarn if free-space below this, emit warning logs
+	SpaceMonThresholdWarn uint64 = 250 * GiB
 
-// monitoring interval
-const spaceMonInterval = 2 * time.Minute
+	// SpaceMonThresholdReadOnly if free-space below this, switch store to read-only
+	SpaceMonThresholdReadOnly uint64 = 200 * GiB
```
danudell-trustnetworks: @kirg Curious as to why do the thresholds need to be so high? Is this a uber use case or does cherami need the diskspace somehow?

PR #340, opened 2017-12-06
Cache results in ReadDestination and ReadConsumerGroup metadata calls…

thuningxu:
… redux

This time, don't cache negative results.

### DIFF
--- a/clients/metadata/metadata_cassandra.go
+++ b/clients/metadata/metadata_cassandra.go
@@ -29,6 +29,7 @@ import (
 	"time"
 
 	"github.com/uber/cherami-server/common"
+	"github.com/uber/cherami-server/common/cache"
 	"github.com/uber/cherami-server/common/configure"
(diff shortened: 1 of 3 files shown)


clients/metadata/metadata_cassandra.go
```diff
@@ -1558,6 +1599,12 @@ func (s *CassandraMetadataService) ReadConsumerGroupByUUID(ctx thrift.Context, r
 		return nil, &shared.BadRequestError{Message: "ConsumerGroupUUID cannot be nil"}
 	}
 
+	key := request.GetConsumerGroupUUID()
```
boy-uber: There are two types of keys here (dstUUID + cgName, ConsumerGroupUUID). They may conflict if cgName is empty. Maybe add some prefix to different type of keys to make sure there is no chance to conflict?

thuningxu: I really hope uuids don't collide...

PR #341, opened 2017-12-11
minor tuning

thuningxu:


kirg:
@thuningxu curious, what are we trying to fix/achieve with this?

PR #344, opened 2018-02-15
Increase KFC num-extents + glide up

kirg:
Increase number of extents created for KFC consumer-groups, from 2 to 4 -- this in turn increases the number of Kafka consumers Cherami would use to consume from the the Kafka topics for the destination.

Also, 'glide up' to get the latest dependencies.


glide.lock
```diff
@@ -136,7 +134,7 @@ imports:
   - common/websocket
   - stream
 - name: github.com/uber/cherami-thrift
-  version: ae427f7974d32a42a96f29f0f627e4d343fad8b2
+  version: 2c02715231abe85d4063130ad3dcf0e44bd33153
```
datoug: this is not the latest version

PR #345, opened 2018-02-15
cmq: lots of various fixes

kirg:
- added new top-level shortcut commands to list extents (x), list destinations (d), cg-extents (cgx), etc
- a new "cmq fix fixdestuuid" command that updates the (newly added) 'destination_uuid' column for each row of 'consumer_groups' table, with the 'consumer_group.destination_uuid' value.
- various minor fixes in 'cmq gc'
  - use "set" package instead of 'map[string]string'
  - removed redundant variables, maps, etc.
- enhanced the 'cmqWriter' interface to make it easier to do "json" output, etc
- implemented 'listStoreExtents' and 'listInputHostExtents' commands
- moved top-level metadata args handling to 'meta.go'
- the delete and undo writers (used by 'cmq gc') creates only one "_delete.cql" and "_undo.cql" each, instead of one per table
- added support to recognize 'status' by numeric values (0 -> open, etc)
- support routines for ANSI control sequences

### DIFF
--- a/cmd/tools/cmq/ansi.go
+++ b/cmd/tools/cmq/ansi.go
@@ -4,49 +4,73 @@ import (
 	"fmt"
 )
 
+// SaveScreen saves the screen
 func SaveScreen() {
 	print(`\x1b[?47h`)
 }
 
+// RestoreScreen restores saved screen
 func RestoreScreen() {
 	print(`\x1b[?47l`)
 }
+
+// ClearScreen clears screen
 func ClearScreen() {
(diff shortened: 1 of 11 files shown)


cmd/tools/cmq/ansi.go
```diff
@@ -0,0 +1,148 @@
...
+	const (
+		start = "\x1b["
+		end   = "m"
+		reset = start + end
+	)
+
+	switch len(t.ansi) {
+	case 0:
+		return t.text
+
```
datoug: can you remove all these extra spaces?

thuningxu:
Can you explain what this diff does?

PR #347, opened 2018-05-09
Add cmq repair for deleting MISSING extents

georgeteo:


cmd/tools/cmq/repair.go
```diff
@@ -0,0 +1,104 @@
...
+
+	deleted := make([]string, 0, len(cqls))
+	defer func() {
+		fmt.Printf("Deleted %d MISSING extents", len(deleted))
+	}()
+	for _, cql := range cqls {
+		ctx, _ := context.WithTimeout(context.Background(), time.Second)
+		if err := mc.session.Query(cql).WithContext(ctx).Exec(); err != nil {
+			fmt.Printf("Error deleting consumer_group_extents: %s\n", err)
+			return
```
kirg: could waiting a 'second' be too short in some scenarios? i might consider adding the timeout as an arg to the command (with a default of one second)  .. so the user is able to override, if necessary.

also, if it times out .. should we perhaps continue with the rest of the them, instead of aborting.

georgeteo: Added timeout flag. 

I think we should abort, because CQL query may return other non-timeout errors and there is no harm in running the command again.

cmd/tools/cmq/repair.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+	}
+}
+
+// generatePlan creates the cql commands for deletion.
+func generatePlan(extentMap map[string]*extentInfo, consumergroupUUID string) []string {
+	var output []string
+	for extentUUID, extentInfo := range extentMap {
+		if extentInfo.extStatus == extMissing {
+			output = append(output, fmt.Sprintf("DELETE FROM consumer_group_extents WHERE consumer_group_uuid=%s AND extent_uuid=%s;", consumergroupUUID, extentUUID))
```
thuningxu: MISSING extents should be deleted from destinations_extents table, too.

cmd/tools/cmq/repair.go
```diff
@@ -0,0 +1,107 @@
...
+	"time"
+)
+
+func repair(c *cli.Context) error {
+	if c.NArg() < 2 {
+		fmt.Printf("dest-uuid and cg-uuid not specified\n")
+		return nil
+	}
+
+	mc, err := newMetadataClient(getOpts(cliContext))
```
thuningxu: Probably should enforce consistency All.

kirg: +1

cmd/tools/cmq/repair.go
kirg: We do **not** have enough information to make a call here .. on whether we can delete an extent or not.

The 'gc' command does all the necessary checks before determining that .. and we can probably just use that for cleaning up any extents that we have missed.

cmd/tools/cmq/repair.go
kirg: +1 .. presuming that would force Cassandra to replicate the delete to all nodes. Downside is the operation might take longer, which I presume is okay?

kirg: Perhaps best to use a default of "All", but leave it 'override-able' with the "-cons' flag to cmq:
```
        mc, err := newMetadataClient(getOpts(cliContext))

        mc.consistency = gocql.All // default to using 'All' consistency
        if cliContext.IsSet("consistency") { // override consistency, if specified
                mc.consistency = gocql.ParseConsistency(cliContext.String("consistency"))
        }
```

thuningxu:
Please also rebase...

cmd/tools/cmq/repair.go
thuningxu: Got it. Please disregard this.

cmd/tools/cmq/repair.go
georgeteo: Just set `gocql.All` for now. I don't anticipate a need to `repair` with lower consistency level.

georgeteo: Never-mind added flag as suggested.

PR #352, opened 2019-02-06
Fix length check.

xiemaisi:
The length of an array is never negative, so the condition as written can never hold.


xiemaisi:
Is there anything I should do to expedite review of this PR? (The CI failure looks unrelated.)

